PR #719, opened 2026-07-01
fix(core): make WSL/remote agent sessions launch correctly

LeTuR:
## Summary

Fixes remote agent sessions running on a WSL distro (or SSH host) that is reached from **inside another WSL distro** — the common thurbox-in-WSL setup. Several failures along the launch path are addressed:

- **`wsl.exe` inherited the caller's cwd** → it tried to `chdir` to a path that does not exist on the target distro, printing `WSL Relay ERROR: chdir(...) failed` and corrupting the tmux control-mode handshake so the agent window never launched. `wsl_command` now passes `--cd /` (a neutral landing dir; every caller overrides the real cwd via `git -C` / tmux `-c`).
- **Agent ran in a non-login shell without `~/.local/bin` on PATH** → `claude` (installed under `~/.local/bin`) was not found, the window command exited 1, and the pane died instantly. The remote window command is now wrapped in a login shell (`/bin/sh -lc 'exec …'`).
- **Hooks `--settings <local path>` did not exist on the remote** → the agent errored `Settings file not found`. thurbox-managed config files referenced in the agent args (under the local config dir) are now materialized on the remote at the same path before launch (transport-neutral, streamed over stdin).
- **Multi-repo symlink workspace was built locally** and its local path handed to the remote. It is now built **on the remote host** (`ensure_remote_workspace`), with the host threaded through `resolve_process_cwd` / `resolve_launch_cwd`; naming mirrors the local builder.
- **Repo picker leaked local paths for a remote target** → local filesystem completion is suppressed for a remote target, and `Tab` now completes against the **remote** filesystem (`list_dir_on`, a variable-free `ls -1p` because `wsl.exe` strips `$var` inside `sh -c`).

## Test plan

- `cargo nextest run --all` — 1686 passing (7 new unit tests: login-shell wrap, remote `sh -c` quoting per transport, variable-free `ls`, workspace-name parity with the local builder, directory-completion suffix logic).
- `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- `cargo test --test architecture_rules` — 12/12.
- Verified end-to-end against a real WSL setup (distro A → distro B): single- and multi-repo sessions launch with Claude Code running in the correct folder, hooks settings materialized, `Tab` completes remote directories.

### DIFF
--- a/src/agent/tmux.rs
+++ b/src/agent/tmux.rs
@@ -682,9 +682,17 @@ impl TmuxBackend {
     }
 
     /// The shell tmux should use for `default-command`. Local uses the user's
-    /// `$SHELL`; a remote backend uses a POSIX shell that is guaranteed to exist
-    /// on the remote host. Not used on Windows (psmux keeps its native default
-    /// shell — see [`apply_session_config`](Self::apply_session_config)).
+    /// `$SHELL`; a remote backend uses a POSIX shell guaranteed to exist on the
+    /// remote host. Not used on Windows (psmux keeps its native default shell —
+    /// see [`apply_session_config`](Self::apply_session_config)).
+    ///
+    /// The value must be a single, space-free token: it round-trips through the
+    /// remote transport's per-argument shell-quoting (`ssh`/`wsl.exe`), where a
+    /// space would be re-split by the remote shell into extra `set-option` args.
+    /// The login-shell `PATH` fix for remote agents (e.g. `claude` under
+    /// `~/.local/bin`) is applied at the *window command* instead — see
+    /// [`build_shell_command`](Self::build_shell_command) /
+    /// [`login_wrap_for_remote`](Self::login_wrap_for_remote).
     #[cfg(not(windows))]
     fn config_shell(&self) -> String {
         if self.transport.is_remote() {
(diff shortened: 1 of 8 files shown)


LeTuR:
Closing as already-merged/superseded.

This work landed on `main` as commit `720db56` ("fix(core): make WSL/remote agent sessions launch correctly" — same title, ~2h after this PR's commit), then was refined by several follow-ups (`e5d2630` dedupe remote helpers, `7330a19` host-adapted args/env, `1d0927b` remote-aware session creation, `90cc6aa` remote hooks status, `1754157` harden spawn arg rewriting, `2af920a` the more robust `--cd /` fix, …).

Rebasing this branch onto `main` conflicts in all 8 files, and every symbol it introduces (`login_wrap_for_remote`, `ensure_remote_workspace`, `list_dir_on`, `host_shell_c`, `dir_completion_suffix`, `remote_path_completion`, `session_process_cwd`, remote file copy, `--cd /`) already exists on `main` in a strictly better form. Resolving the conflicts would produce an essentially empty commit, so there's nothing left to merge here.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #809, opened 2026-07-15
feat(core): hooks-driven status for all agents on remote hosts

LeTuR:
## Summary
- Remote sessions now report hooks-driven status for **every** agent, not just claude: `session_ops::remote_hooks` provisions each config-dir agent's hook payload (codex, antigravity, opencode, vibe, copilot) onto the host at spawn time, with the local installer's safety rules — `requires_dir` probed over ssh, prune-then-merge for shared JSON (both `session signal` and `@thurbox_state` markers, so upgrades replace instead of accumulate), managed-marker guard for standalone files, compare-before-write, cached per `(backend, agent)`, best-effort (never fails a spawn).
- The remote signal rewrite now maps over **every launch arg**, fixing aider's literal `--notifications-command "thurbox-cli …"` silently no-oping on hosts.
- Mux-aware rewrite target: psmux hosts get a socket-explicit `psmux -L <socket> set-option -p @thurbox_state` form (psmux has no in-pane `$TMUX` socket resolution).
- psmux hosts get a **status channel**: psmux has no format subscriptions, so `ControlMode` runs a 1 s pane-option poller (`list-panes -F` diffed by the pure `diff_polled_hook_states`) feeding the same `sub_events` queue — everything downstream of `drain_remote_hook_events` is shared and unchanged. The poller (an active 1 Hz command, unlike the passive tmux subscription) arms only on **remote** psmux connections and only once the gate below is flipped — a local psmux (Windows) session signals via `thurbox-cli` directly and never sets the pane option.
- the whole psmux path — hook **shipping** and the **poller** — stays gated off on one switch (`session::psmux_hook_rewrite_supported()` = false) until `windows-vm.sh test`'s new probes prove the pane-user-option behavior against the pinned psmux; the Windows-copy helpers (`remote_home_windows`, base64 PowerShell file write) land ready behind the gate.
- **Status flows with the TUI closed too**: the live channels above die with the TUI's control-mode connection, so the headless `automation tick` (the 60 s heartbeat keeper) now polls each host that has live remote sessions (`remote_hooks::poll_remote_hook_states` — one-shot `list-panes -F`, allow-listed, diffed against the stored `hook_state`, so an acknowledged `done` is never resurrected and no new table is needed) and writes changes into the same columns; `session list`/`get` JSON now carries the raw `hook_state` so headless consumers can actually read it.
- **Degradation is visible**: stripped/failed hook wiring lands on `SessionInfo.hook_wiring` and renders as a `Hooks: degraded — <reason>` row in the info panel instead of only a log warn.
- Latent freeze fix: the interactive spawn's remote arg adaptation (ssh round-trips) moves off the UI thread into the spawn worker, alongside `ensure_backend_ready` (ADR-P12).

## Test plan
- `cargo nextest run --all` — 1848 pass, incl. new unit tests for the mux-aware rewrite, all-assets marker guard, merge idempotency/upgrade-safety, poll diff semantics, and the info-panel row
- `cargo clippy --all-targets --all-features -- -D warnings`, rustdoc, `cargo deny` (new direct dep `base64`, already in the tree transitively), architecture rules — all clean
- **Live e2e**: `scripts/dev/e2e/linux-container.sh test` (extended) — verified against a real sshd+tmux container: codex `hooks.json` lands rewritten in `~/.codex`, no `thurbox-cli` reference survives, byte-stable across a second spawn (cross-process merge idempotency), arg-carried config materializes rewritten, and the headless poll round-trips (pane option set in-pane-style → `automation tick` → `hook_state` in the CLI JSON; second tick a stable no-op) — PASS. The e2e sandbox now pins `THURBOX_CONFIG_DIR`/`THURBOX_DATA_DIR` so running it from inside a thurbox session can't leak onto the real config/DB
- `scripts/dev/e2e/windows-vm.sh test` gains gate-evidence probes A/B (pane user options in `list-panes -F`; id-less in-pane `set-option -p`) — not run here (needs KVM + Windows VM); the gate stays closed so no behavior depends on them yet

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -694,16 +694,20 @@ not block startup, so `check_available`/`ensure_ready` are deferred to first use
   connection (`refresh-client -B 'thurbox-status:%*:#{@thurbox_state}'`,
   armed in `ControlMode::start` so reconnects re-arm; tmux ≥ 3.2 = the
   existing floor) and receives `%subscription-changed` pushes (≤1/s); a
-  **psmux** connection instead runs a 1 s **poller thread** (`list-panes -F`
-  diffed by `control_mode::diff_polled_hook_states`) feeding the same queue. Both
+  **remote psmux** connection instead runs a 1 s **poller thread**
+  (`list-panes -F` diffed by `control_mode::diff_polled_hook_states`) feeding
+  the same queue — armed only behind the psmux gate below (a poll is an
+  active per-second command, unlike the passive subscription, and a *local*
+  psmux session signals via `thurbox-cli` directly). Both
   channels drain each tick via `App::drain_remote_hook_events` into the same
   `set_hook_state` columns local signals use — so Done→seen acknowledgment,
   OS notifications, and the stuck-`working` fallback are shared. Events are
   matched by **backend name + pane id** (pane ids collide across hosts),
   allow-listed (remote-controlled text), and deduped against the cache (a
   reconnect re-report must not resurrect an acknowledged `done`). Remaining
-  carve-out: hook **provisioning onto psmux/Windows hosts** is gated off
-  (`spawn::psmux_hook_rewrite_supported`) until the psmux behaviors are
+  carve-out: the **whole psmux/Windows-host path** — hook provisioning,
+  rewrite shipping, and the status poller — is gated off on the one switch
+  (`session::psmux_hook_rewrite_supported`) until the psmux behaviors are
   proven by `scripts/dev/e2e/windows-vm.sh test`'s probes — such sessions
   show a `Hooks: degraded` hint instead of silently idling.
 - **Remote teardown** (WSL inherits the SSH path): `session delete --force`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -662,33 +662,50 @@ not block startup, so `check_available`/`ensure_ready` are deferred to first use
 - **Agent config on the host**: agent args referencing thurbox-managed config
   by *local* path (the hooks extension's `--settings <config>/hooks/
   claude.json`) would kill the remote agent on launch ("Settings file not
-  found"). `session_ops::spawn::adapt_agent_args_for_remote` (shared by
-  headless spawn and the TUI) rewrites them per host: on a POSIX remote the
-  home-anchored path is **translated to the remote home**, the file copied
-  there, and the arg substituted; on a psmux host / non-POSIX config root /
-  failed copy the **flag+path pair is stripped** so the agent launches clean.
+  found"). `session_ops::spawn::adapt_def_for_launch` (shared by headless
+  spawn and the TUI, run on the spawn worker — never the UI thread) rewrites
+  them per host: on a POSIX remote the home-anchored path is **translated to
+  the remote home**, the file copied there, and the arg substituted; on a
+  psmux host (while `psmux_hook_rewrite_supported` stays off) / non-POSIX
+  config root / failed copy the **flag+path pair is stripped** so the agent
+  launches clean — surfaced as a `Hooks: degraded` row in the info panel
+  (`SessionInfo.hook_wiring`). Literal signal commands carried directly in
+  args (aider's `--notifications-command`) are rewritten too.
   The local-path env hints
   (`THURBOX_METRICS_DIR`/`THURBOX_CONFIG_DIR`/`THURBOX_DATA_DIR`) are likewise
   skipped for remote spawns (`inject_thurbox_env`); only the opaque identity
   vars travel.
-- **Remote session status** (hooks-driven, like local): `thurbox-cli session
-  signal` can't work from a host (no CLI there; it would write the host's own
-  DB), so the materialized hook file's commands are **rewritten**
-  (`builtin_hooks::rewrite_hook_signals_for_remote`) to set a tmux **pane user
+- **Remote session status** (hooks-driven, like local, **all agents**):
+  `thurbox-cli session signal` can't work from a host (no CLI there; it would
+  write the host's own DB), so hook commands are **rewritten**
+  (`builtin_hooks::rewrite_hook_signals_for_target`) to set a tmux **pane user
   option** instead — `tmux set-option -p @thurbox_state <s>` needs no socket,
-  pane id, or identity inside a pane. The local TUI's persistent control-mode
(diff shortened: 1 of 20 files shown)


LeTuR:
Reviewed and pushed review follow-ups in 99ef114:

**Fixed**
- **Poller gating** — the psmux pane-option poller was armed on *every* psmux connection (including local Windows, where sessions signal via `thurbox-cli` directly), and while the shipping gate is closed no host has a producer, so it was a guaranteed-empty 1 Hz command forever. It now arms only for **remote** psmux connections behind the same gate. The gate moved to `session::psmux_hook_rewrite_supported` (pure-data layer, arch-rule-clean for both `agent` and `session_ops`) so flipping it enables the whole psmux path — shipping *and* status channel — in one switch.
- **Provisioning lock scope** — the process-wide cache mutex was held across the ssh probe/read/merge/write, so a slow or down host serialized provisioning for *unrelated* hosts. Now: short lock for check + per-key in-flight guard, ssh work unlocked, re-lock to record. Same-key merges stay exclusive.
- **Cache semantics** — an "agent not installed" probe result was cached as provisioned, contradicting the "failures are retried" doc: installing the agent on the host later would silently never get hooks until restart. Tri-state `ProvisionOutcome`; only `Provisioned` is cached, `NotInstalled` re-probes each spawn (one cheap round-trip).
- **Socket sanitization** — the psmux rewrite spliced the user-authored `hosts.toml` socket name into JSON/JS/TOML by byte-level replace while a comment claimed it could only be `thurbox`/`thurbox-dev`. Now sanitized to `[A-Za-z0-9._-]` at target construction (warn + degrade-dark rather than corrupt the shipped config), with a test.
- A poisoned provisioning-cache lock now reports a degradation instead of silently claiming healthy wiring.

**Noted, deliberately not changed here**
- Restart/respawn (`launch_provider_for`) still skips provisioning and drops the degradation note — fine for the common already-provisioned case; worth folding into `adapt_def_for_launch` in a follow-up.
- `hook_wiring` is transient across TUI restarts while the condition persists — v1-acceptable.
- Headless status freeze while no TUI is open (the control-mode channel dies with the App) — separate follow-up; the heartbeat `automation tick` polling DB-diffed pane state is the likely shape.

All 16 pre-commit hooks green; 1846 tests pass; docs (CLAUDE.md/FEATURES.md/hooks README/windows-vm.sh) updated for the moved gate + poller gating.

LeTuR:
be87d85 closes the last identified gap: **remote status froze whenever no TUI was attached** (the subscription/psmux channels live on the TUI's control-mode connection; local sessions were immune because `session signal` writes SQLite directly).

**Shape**
- The headless `automation tick` (60 s heartbeat keeper, already running with the TUI closed) polls each host that has **live remote sessions in the DB** — hosts.toml entries with no sessions are never contacted — via a one-shot read-only `list-panes -F` (`agent::tmux::list_remote_hook_states`, deliberately no `ensure_ready` so a poll can't create the remote server).
- **No new last-seen table**: polled values are allow-listed and diffed against the stored `hook_state` — the DB *is* the last-seen store, and an acknowledged `done` (`seen_at` is a separate column) compares equal and is never resurrected. Parsing is the shared `control_mode::parse_pane_hook_states`, extracted from the psmux poller's diff.
- `session list`/`get` JSON now carries the raw `hook_state` (`null` = never reported) — the headless consumer this feeds; without it the polled states were invisible until a TUI opened. The TUI stays the sub-second channel while open; 60 s is fine for script readers.
- psmux hosts stay behind `session::psmux_hook_rewrite_supported`; unreachable hosts are skipped per cycle (ConnectTimeout-bounded, best-effort); skipped entirely when the hooks extension is opted out.

**Verification**
- 1848 tests pass (new pure-fn tests: changes-only writes, acknowledged-done non-resurrection, state allow-listing).
- **Live e2e** against the real sshd+tmux container: pane option set in-pane-style → `automation tick` → `hook_state` lands in the CLI JSON; a second tick is a stable no-op — PASS.
- Drive-by hermeticity fix surfaced by that run: the e2e now pins `THURBOX_CONFIG_DIR`/`THURBOX_DATA_DIR` to its sandbox — running it from *inside* a thurbox session injects those overrides, which outrank `XDG_*` and pointed the CLI at the real hosts.toml/DB.
- All 16 pre-commit hooks green; docs updated (CLAUDE.md, FEATURES.md, hooks README).

LeTuR:
Adversarial correctness review (multi-agent, 4 finder angles) + fixes pushed as `3561888`, and the branch is rebased onto latest main (the repo-picker work moved `list_dir_on`/`list_dir_command` → `list_dir_entries_on`; resolved by dropping the superseded pair). Auto-merge is enabled.

Verified findings fixed:
- **`send_command_on` lock scope** — the response-queue lock was held across the blocking pipe write, so a wedged transport could stall the reader thread's response dispatch (and with it every pane's output for that host). Now stdin is locked first (still serializing enqueue+write, so FIFO waiter order matches wire order) and the queue lock is only taken for the brief push/pop.
- **Concurrent same-`(host, agent)` provisioning** — the in-flight skip reported the second session healthy while its agent could boot before (or after a *failed*) first pass. The waiter now blocks (bounded) and retries the pass itself if the holder failed; a Drop guard releases the key even on panic, and cache access recovers from mutex poisoning (all mutations are single atomic ops — and the guard's own unwind-time release would otherwise poison it).
- **`read_remote_file` misdiagnosis** — `cat`'s own exit 1 (e.g. permission denied) was reported as 'exists but is not a regular file'; a distinct sentinel now separates the two and surfaces cat's stderr.
- **psmux poller session-name injection** — the hosts.toml `session` value was embedded unquoted in the wire command; now double-quote-framed + sanitized like the socket.
- **Allow-list drift** — the hook-state list was duplicated in the CLI parser, the TUI drain, and the headless poll; hoisted to `session::HOOK_STATES`.
- **Drift-guard hole** — the asset test now keys on `thurbox-cli session signal --…`, catching a command whose flags drift (e.g. an inserted `--quiet`) that escaped both sides of the old count.
- **Tick ordering** — the headless status poll now runs *after* the due-automation pass, so an unreachable host's ConnectTimeout can't delay a scheduled firing.

1866 tests pass post-rebase; live container e2e (provisioning + headless poll round-trip) PASS; all pre-commit hooks green; all four branch commits signed.

PR #922, opened 2026-08-11
fix(core): type hyphens on Windows through psmux's send-keys parser

### LINKED ISSUE #920: Impossible to write hyphens
`-` do not get printed in Thurbox.

Whereas `_` work correctly.

##  Environment

Thurbox version: v1.7.0

Powershell + local session + claude

LeTuR:
## Summary

- Fixes #920 — a `-` typed into a session on Windows never reached the pane, while `_` and every other character did.
- Root cause, from psmux 3.3.6's own source (`src/server/connection.rs`): the control-mode `send-keys` handler classifies arguments **after** tokenizing, and `parse_command_line` strips the quotes first. So `send-keys -t %1 -l -N 1 "-"` arrived as the bare argument `-` and its `!a.starts_with('-')` filter dropped it as an unknown flag. Quoting could never have protected against this, so the claim that it did is gone from `psmux_quote`'s doc.
- `psmux_literal_args` re-emits the offending leading character as psmux's own `0xNN` codepoint argument, which psmux decodes back to the same character and — in literal mode — joins with no separator, so the run is reassembled exactly. Escaping repeats until the remainder is safe (`--x` needs both hyphens; `-0x41` the hyphen and then the `0`). Interior hyphens are untouched.
- Closes a sibling bug of the same class found while in that handler: psmux rewrites **any** `0xNN`-shaped argument into the character it names, so a literal run spelling `0x41` would have been typed as `A`. The same escape covers it.
- Docs: updated the psmux-divergences bullet in `CLAUDE.md` per the repo's same-PR doc rule.

## Test plan

- `cargo nextest run -E 'test(psmux)'` — 23/23 pass, including 3 new tests: leading-hyphen escaping (`-`, `--flag=a-b`), the `0xNN` lookalike (`0x41`, `-0x41`, and the non-lookalike `0x41z`), and the all-hyphen/empty run.
- `cargo clippy --all-targets --all-features -- -D warnings` and `shellcheck` clean.
- Probe D added to `scripts/dev/e2e/windows-vm.sh test`: types both encodings into one prompt line on a real psmux and asserts the old form loses its hyphen while the `0xNN` form keeps it. Evidence-level like probes A–C, so a future psmux fix surfaces as a reported observation rather than silence.
- Not run: the Windows VM suite (needs `/dev/kvm`); CI's native `windows` job covers it.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -712,6 +712,15 @@ control-mode protocol is byte-identical over either transport/binary, with
   `'` typed arrived in the pane as `\` (`it's` → `it\s`) — the `-N` flag makes
   psmux's coalescing decoder bail, and its direct handler reads the
   double-quote framing correctly (`flush_psmux_literal` / `psmux_quote`).
+  Quoting alone can't carry every run, because psmux classifies arguments
+  *after* tokenizing (the quotes are gone by then): it drops each one starting
+  with `-` as an unknown flag, so a typed hyphen never reached the pane (issue
+  #920), and it rewrites a `0xNN`-shaped argument into the character it names,
+  so a run spelling `0x41` would arrive as `A`. `psmux_literal_args` escapes
+  both by re-emitting the offending leading character *as* a `0xNN` argument —
+  psmux decodes it back and, in literal mode, joins the arguments with no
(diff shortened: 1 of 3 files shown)


pleymor:
Now works like a charm <3

PR #1004, opened 2026-08-25
fix(core): give local spawns a pane id so duplicate names attach

rchoffar:
## Summary
- A local spawn stored no pane id and the TUI resolved its pane by tmux window name (`tb-<name>`), which is not unique: creating two sessions with the same name (accepting the creation flow's proposed default twice) left the second stuck at "no live terminal" — the resolver deliberately refuses an ambiguous window name — and name-targeted lifecycle ops (restart/delete kill, capture, prompt-send) could hit the wrong duplicate's window.
- Local spawns now learn their pane id up front (`new-window -P`, matching what remote spawns already did) and persist it; the one-shot tmux helpers target the pane id first, verifying it still sits in the session's own `tb-` window (tmux reuses pane numbers across server restarts), with the window name kept as fallback for legacy rows.
- Legacy rows migrate themselves: a pane the interface still resolved by name has its id written back after the first successful adoption (`Database::set_backend_id`, drained by the loop). Windows/psmux keeps today's empty-id behavior until `-P` is probed there.
- Also actually applies `tests/create_e2e.rs`'s throwaway tmux socket — it was aspirational, so its spawned windows landed on (and raced) the real `thurbox-dev` server.

## Test plan
- [x] `cargo nextest run --all` — 1944 pass (plus clippy `-D warnings`, `cargo doc -D warnings`, fmt, architecture rules)
- [x] New test: two same-named sessions with distinct pane ids both attach (`tests/v2_attach_by_name.rs`)
- [x] New test: two same-named spawns get distinct pane ids (`tests/create_e2e.rs`)
- [x] Sandbox repro: `session create --name thurbox` twice on one repo → rows persisted with `%1`/`%2`
- [ ] Manual TUI check: Ctrl+N twice on the same repo accepting the default name → both panes live; Ctrl+D one → the other survives

### DIFF
--- a/src/agent/tmux.rs
+++ b/src/agent/tmux.rs
@@ -158,6 +158,34 @@ fn window_target(session_name: &str) -> String {
     format!("{TMUX_SESSION}:={}", agent_window_name(session_name))
 }
 
+/// Whether `pane_id` (`%N`) is alive and sits in a window named `window`.
+///
+/// The verification is what makes a *persisted* pane id safe to target: tmux
+/// reuses pane numbers after a server restart, so a stored id can point at a
+/// different window entirely (a shell, the heartbeat keeper). Checking the
+/// window name catches that; a dead or reassigned id falls back to the name.
+fn pane_matches_window(pane_id: &str, window: &str) -> bool {
+    let out =
+        local_mux_command(&["display-message", "-p", "-t", pane_id, "#{window_name}"]).output();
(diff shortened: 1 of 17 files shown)


LeTuR (approved):
Reviewed — the pane-id-first resolution with the `tb-` window verification and the legacy name fallback is sound, and the e2e socket fix removes a real race. Approving.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1023, opened 2026-08-28
feat(core): select and focus a newly created session at submit

rchoffar:
## Summary
- Creating a session (Ctrl+N flow) or forking one (Ctrl+F) now selects the new session and moves keyboard focus to the agent pane **at submit time**: the loop mints the `SessionId` in `dispatch_tracked` and steers the view onto it before handing the command to the worker, and the session list's sticky follow puts the cursor on the row the moment the spawn lands it.
- This deliberately reverses the documented "creating a session moves nothing" decision while keeping its rationale: the objection was to *completion-time* steering (the spawn finishes on a worker seconds after the wizard closed). Submit time is a moment the user chose; completion still moves nothing. The old-decision comments/docs (`focus_on_session`, `execute.rs`, `poll_command_bus`, `docs/FEATURES.md`) are rewritten to record the new line.
- No Lua changes: the follow mechanism was already sticky-until-found — a failed spawn's dangling follow is harmless, the user's own cursor move mid-spawn wins, several rapid creates follow only the last id, and headless creates (`thurbox-cli`, automations, task dispatch) never pass through the mint so they keep moving nothing.
- Rides along: `test(core)` portability fix for `tests/tui_e2e.rs` — its `libc::openpty` call used Linux's `*const` signature, which does not compile on macOS and blocked every local pre-commit run there.

## Test plan
- [x] `cargo nextest run --all` (1950+ tests, via pre-commit hooks on both commits)
- [x] New `v2_session_list` tests pin the load-bearing semantics: a focus request for an id not yet in the snapshot waits (sticky follow), and a user cursor move mid-wait wins over the pending jump
- [x] New `mint_session_id` unit test: stamps `Create`/`Fork`, returns `None` for everything else; `create_e2e` asserts the spawn echoes a pre-minted id
- [ ] Manual sandbox check (`scripts/dev/sandbox.sh`): Ctrl+N through the wizard → focus lands on the agent pane at Enter, cursor lands on the new row when it appears; same for Ctrl+F; `thurbox-cli session create` moves nothing

### DIFF
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -247,15 +247,20 @@ not applicable.
    session. Skipped when only one agent is defined in
    `agents.toml`.
 
-**Creating a session moves nothing.** The new row appears in the list and waits
-to be picked; the selection, the pane showing it and the keyboard all stay where
-they were. Creation is a command that finishes on a worker seconds after the flow
-closed, so the moment it lands is not a moment the user chose — steering the view
-then interrupted whatever they had gone back to reading, and made creating three
-sessions in a row a fight with the cursor. `Ctrl+F` fork behaves the same way.
-Selection is still *steerable*, by the two requests that are deliberate: a
-clicked notification and `thurbox-cli session focus`, both through
-`focus_session`, which the list follows by id rather than by row number.
+**Creating a session selects it — at submit, not at completion.** Confirming
+the flow (and `Ctrl+F` fork) mints the new session's id on the spot, moves the
+keyboard to the agent pane at once, and sets the list *following* that id
+(`focus_session`, the same channel a clicked notification and `thurbox-cli
+session focus` use — by id, never by row number). The spawn itself still
+finishes on a worker seconds later, and that landing moves nothing: the follow
+is sticky until the row appears, so the cursor is already waiting on it — while
+completion-time steering yanked the view at a moment the user did not choose.
+Until the row lands the agent pane keeps showing the previous session; moving
+the cursor yourself in the meantime wins over the pending jump (and clears a
+follow left dangling by a failed spawn); creating several sessions in a row
+follows only the last one. Headless creates (`thurbox-cli session create`,
+automations, task dispatch) still move nothing — only `session focus` steers
+cross-process.
 
 A session is fully described by its repos and agent. There is no
 per-session model selection, permissions, prompt, tool, or skill
--- a/src/coordinator/commands.rs
+++ b/src/coordinator/commands.rs
@@ -226,9 +226,10 @@ impl App {
             // the command made, which only exists here once the rows are.
             self.snapshots.refresh();
             self.report_finished_commands();
-            // A finished creation is deliberately not acted on here: see
-            // `focus_on_session` for why a session that just spawned does not
-            // pull the view onto itself.
+            // A finished creation moves nothing here: the view started
+            // following its pre-minted id back at dispatch (`dispatch_tracked`)
+            // — the moment the user chose — so by the time the row lands the
+            // cursor is already waiting on it.
             self.note_data_change();
         } else {
             self.snapshots.refresh_if_due();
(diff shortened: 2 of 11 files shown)


LeTuR (changes requested):
Thanks for the careful write-up — the submit-time vs completion-time distinction is a real refinement of the argument, and the mechanism (pre-mint at dispatch, sticky follow by id) is the right shape. But the answer on the default is no: **creating a session should keep moving nothing.**

### 1. This reverses a decision, not an oversight

#988 (`fix(ui): stop pulling the view onto a session that just finished spawning`) removed exactly this behaviour deliberately, and `docs/FEATURES.md` records it as a decision with its rationale. Rewriting that paragraph to say the opposite is the part I want to push back on hardest — the documented line is the product decision, and it stands.

Moving the trigger from completion to submit answers *one* of #988's two objections (the "moment nobody chose" half). It does not answer the other: creating several sessions in a row still means the view walks away from whatever you were reading, and the common case for `Ctrl+N` is *queueing up work you are not about to babysit*.

**If we want this at all, it has to be opt-in and off by default:** a core flag (`[features] focus_new_session = false` in `settings.toml`, live-appliable, surfaced in the settings panel) or a `10_sessions` plugin setting, checked in `dispatch_tracked` before the `focus_on_session` call. `FEATURES.md` then keeps today's rule as the documented default and describes the flag as the opt-out from it. Headless creates stay untouched either way, as you already have them.

### 2. A concrete bug, independent of the taste question

Submit-time focus moves the **keyboard** onto the agent pane while `store.selected` is still the *previous* session — the PR body notes this ("until the row lands the agent pane keeps showing the previous session") but treats it as cosmetic. It isn't:

- `20_agent.lua` renders its surface from `store.selected`;
- `coordinator::draw` stamps `focused_surface` from the surface the focused pane actually rendered (`draw.rs:486`);
- `coordinator::input` routes raw keys straight to `focused_surface` (`input.rs:360`).

So between `Enter` on the wizard and the row landing, **every keystroke is typed into the previously selected agent's prompt.** That window is a worktree create plus a tmux spawn — seconds locally, and considerably longer for an `ssh:`/`wsl:` host, which is where `ensure_ready` can run out a whole ssh timeout — and it is precisely the moment a user starts typing their first prompt. Completion-time focus never had this hazard, because the row existed before the keyboard moved.

If the flag ships, focus has to be split from the follow: let the *selection* follow the pre-minted id (sticky, as you have it), but move the **input focus** only when the row actually lands — or have the agent pane draw a placeholder for a pending id rather than the old session's live terminal.

### 3. A failed spawn leaves the keyboard moved

The dangling `state.follow` really is harmless for the selection, as you say. The focus ring is not: `focus_on_session` has already set `self.focus` to the agent pane, and nothing puts it back when the create errors. So a rejected create (pre-create hook veto, bad branch, unreachable host) leaves the user focused on — and typing into — an unrelated session's terminal, with the error in the panel behind them.

### 4. Please split the ride-along

The `tests/tui_e2e.rs` `openpty` portability fix is unrelated and blocks macOS pre-commit runs *today*. Send it as its own `test(core)` PR and it can land immediately regardless of how this discussion lands. The reasoning in the comment (raw `*mut` local to satisfy both signatures without tripping `unnecessary_mut_passed`) is right.

### What is worth keeping either way

- `SpawnRequest.session_id` and `Command::mint_session_id` are a clean mechanism — a caller that must know the id before the worker finishes is a legitimate need, and `parse` staying pure is the right call.
- The two new `v2_session_list` tests pin sticky-follow semantics that a clicked notification and `thurbox-cli session focus` already depend on and nothing pinned before. Those are worth having with the default off.

So: keep the machinery, put the behaviour behind a flag that defaults to off, fix the keyboard-lands-on-the-old-session window before the flag is usable at all, and leave `FEATURES.md`'s current rule standing as the default with the flag documented beside it.

LeTuR:
Superseded by #1025, so I'm closing this one.

Same behaviour, but **opt-in and Lua-only**. #1025 adds a `focus_new_session` setting to the session list (settings → Sessions, off by default); turned on, a create or a fork selects the new session and gives the agent pane the keyboard, exactly as `Enter` on its row would. Turned off — the default — creating a session still moves nothing, so the decision this PR reversed stands for anyone who hasn't asked otherwise.

It needs no kernel change at all: the list already owns the selection and the sticky follow, so it just subscribes to `session.post_create` — an event that already fires only for a create *this interface* performed, which gives the "headless creates never steal focus" property for free.

The one thing it does **not** do is the part of this PR that genuinely needed Rust: steering at **submit** rather than at **completion**. The id does not exist until the worker mints it, so submit-time focus still needs `Command::mint_session_id`. If that's wanted, the clean follow-up is to bring that half back with the setting as its gate (have `dispatch_tracked` read it) rather than unconditionally — otherwise both jumps are live at once and the switch reads as broken when it is off.

The `test(core)` macOS `openpty` portability fix riding along here is worth keeping either way; it's not in #1025, so it wants its own small PR.

PR #1042, opened 2026-08-31
feat: open an existing worktree instead of always creating one

EmilSkaaning:
## What

The new-session flow could only ever **create** a worktree, at thurbox's derived
path, via `git worktree add -b`. So a worktree made outside thurbox —
`.worktrees/`, a sibling directory, anywhere — was invisible (existence was only
probed by building the derived path), and a branch already checked out could not
be opened at all.

Now a repo's existing worktrees appear as `↳` child rows under whichever repo the
picker's cursor is on, and `Enter` opens one with **no prompts**: no base branch,
no session name, no branch name.

## How

- **`git::list_worktrees_on`** — `git worktree list --porcelain`, so worktrees are
  found wherever they live. Routed through the existing `git_command(host, …)`
  helper, so remote hosts work unchanged. Drops the main checkout, bare repos,
  detached heads and prunable registrations; keeps `locked`.
- **`kernel::repos`** — a fourth parameterised read (`want_worktrees` →
  `thurbox.worktrees`). Kept separate from `Branches`: the picker needs it before
  the branch step exists, and the two answer different questions. No `git fetch` —
  local state, read on a keypress. One call per highlighted row, cached under the
  branch-list TTL.
- **`session_ops::spawn`** — `SpawnRequest::existing_worktree` opens rather than
  creates: no `git worktree add`, the path becomes the cwd, recorded at worktree 0
  ahead of any extra repos.
- **Naming** — an opened worktree names the session after the worktree *directory*,
  falling back to the branch. `.worktrees/optimize-dot-products` on
  `bolt/optimize-dot-products-6492608039562866937` gives a session called
  `optimize-dot-products`. Creating a worktree is unchanged.

- **Scope** — opening is TUI-only for now: `thurbox-cli session create` has no
  `--worktree-path`, so the kernel `create` command is the only caller that can
  pass one. Creating a worktree from the CLI is unchanged.

## Review catch

`resolve_dirs` keyed the primary cwd off `worktree_branch` rather than "is a
worktree being opened", so a `create` carrying `worktree_path` without `branch`
put the session in the repo root while recording a worktree elsewhere.
Unreachable from the shipped Lua, reachable from any plugin. Fixed, with a
regression test that fails on the pre-fix code; `create` now refuses the half-pair.

## Test plan

- [x] `cargo nextest run --lib` — 1612 pass; integration binaries — 307 pass
- [x] `cargo fmt --check`, `clippy -D warnings` (lib+bins), `cargo doc -D warnings`
- [x] `selene ui`, `stylua --check ui`, `rumdl check`
- [x] **Manual** — opened `.worktrees/optimize-dot-products` in the TUI: child rows
      render with branch muted, no prompts, session named after the directory, and
      `git worktree list` unchanged
- [x] `lua-language-server --check ui --configpath "$PWD/.luarc.json" --checklevel=Warning`
      — no problems found

New coverage: porcelain parsing (linked-only, `locked` kept, missing trailing
newline, spaces in paths), `resolve_dirs` opening without registering a worktree
and leading its extras, `session_name` precedence, the full Lua flow asserting the
exact `Command::Create`, and a tmux e2e opening a worktree at a foreign path.

**Pre-existing, not from this branch:** `tests/tui_e2e.rs` does not compile against
libc 0.2.189 on this toolchain, and three `session_ops::shared_tests` fail under
plain `cargo test` (they need nextest's process isolation). Both reproduce on
clean `main`.

Screenshot 2026-08-31 at 12 19 20


EmilSkaaning:
Closing while I validate this locally in the TUI — will reopen once the flow is confirmed by hand. Apologies for the early noise.

LeTuR:
Reviewed as LeTuR's agent, against `main` at `dacee109` (#1044, merged 2026-09-01).
Test-merged in a scratch worktree; the merged tree type-checks. I did not run the
suite.

#1044 named this work as deferred to its own PR, so the overlap is expected — but it
landed first and takes v41.

## Rebase

- Schema version collision. #1044 shipped `migrate_v41_joinable`; this becomes **v42**.
  The number also appears in `sync/state.rs` (`created_by_thurbox` doc) and
  `docs/FEATURES.md`. The `SCHEMA_VERSION` header comment gained no v41 clause — every
  prior version has one.
- One compile break: `tests/create_e2e.rs:868` (#1044's `recipe-probe`) is a full struct
  literal and needs `existing_worktree: None`. The only one — `SpawnRequest` derives
  `Default`.

## `kept_worktrees` has no reader

The field is written and never read. Three places need it, not one:

- `src/cli/sessions.rs:909` — the delete JSON
- `src/cli/sessions.rs:928` — `force_delete_detail`, the human output
- `src/session_ops/delete.rs` — the ADR-24 delegated path parses `removed_worktrees`,
  `worktree_errors` and `remote_teardown_error` out of the host CLI's answer. Fix only
  the local output and a shared-host delete still reports nothing.

Today `session delete --force` on a session that opened someone's worktree prints
`removed worktrees: 0` and stops. That is the silence the field's own doc says it
exists to prevent, and `docs/FEATURES.md` already claims it is reported. #1044 landed
`tests/cli_stdout_contract.rs` pinning this output, so it wants a case.

## The restore refusal is now wrong for a borrowed worktree

`src/session_ops/restore.rs:57` gates on `deleted.force_deleted` alone:

> was force-deleted; recovering it brings back committed work only (uncommitted and
> untracked changes are gone)

Force-delete a session whose worktrees were all borrowed and nothing on disk was
touched. The message still fires, and it is false — the work is still in the directory.
The gate wants to key on whether anything was actually removed.

## Session names collide with #1044's resolver

The naming change is the selling point: `feat/dynamic-tooltips-15307729713678226529`
becomes `dynamic-tooltips`. That drops what made auto-generated names unique.

#1044 routed every session verb through `session_ref::resolve` — 20 call sites in
`src/cli/sessions.rs` — and its contract is "ambiguity is an error, never a guess".
Kernel `create` builds `SpawnRequest` directly, so nothing dedupes.

```mermaid
flowchart LR
    A["repoA/.worktrees/tooltips"] --> N["session 'tooltips'"]
    B["repoB/.worktrees/tooltips"] --> M["session 'tooltips'"]
    N --> R{"thurbox-cli send tooltips"}
    M --> R
    R --> E["refused: ambiguous"]
```

Both sessions become unreachable by name from the CLI — `send`, `key`, `capture`,
`stop`, `exec`, `meta`, `delete`, `watch`. Before #1044 that was cosmetic. A
uniquifying fallback in `session_name` would settle it.

## Comments the change invalidated

`645e8c2a` fixed `docs/PLUGINS.md` for the fourth read and missed three:

- `src/kernel/host/publish.rs:366` — headline says "four parameterised reads", the next
  sentence still says "so all three are empty tables"
- `src/kernel/repos.rs:275` — "the creation flow's three reads"
- `src/coordinator/mod.rs:326` — "costs three lookups"

Separately, `src/kernel/command/execute.rs`: the `session_name` doc block sits directly
under `create`'s with no blank line, so "The whole pipeline ... is what `thurbox-cli
session create` uses" now documents a string helper and `fn create` has no doc.

## nit

- `recreate_worktrees` guards the thurbox arm on `branch_exists` and pushes the borrowed
  arm back unchecked. If the directory is gone, the restored session gets a cwd that
  is not there. `worktree_path.is_dir()` makes the arms symmetric.
- `create` runs `expand_tilde` with the local home even when `host` is set, and skips the
  `is_dir` check for remotes. Picker-safe; plugin-reachable.
- No `--worktree-path` on `thurbox-cli session create`, so this is TUI-only. Fine to
  defer, worth saying so in the description.

## Checked and correct

`parse_worktree_list` (stanza counter, trailing flush, all four drop cases, spaces in
paths, `prunable <reason>`). The `session_from_json` skew default. The publish memo —
`repos.poll()` bumps `epoch.data`, so Pending→Ready repaints. And the restore
short-circuit is not just an optimisation: `add_existing_worktree` derives thurbox's own
path, so without it a borrowed worktree would fail `git worktree add` and get dropped by
the `warn!` arm.

LeTuR:
Correction to my review above: withdraw the "session names collide with #1044's
resolver" section. Ignore it.

`SessionId` is a stable unique UUID and the DB primary key; names were never unique,
and three places say so outright — `storage/sessions.rs:387`, `cli/session_ref.rs:16`,
`cli/sessions.rs:1349`. #1044's resolver refusing an ambiguous name is the designed
answer to a case ADR-24 already guarantees (a mirrored host contributes rows that
legitimately collide), not something this PR breaks. Sessions stay addressable by UUID
or id prefix either way.

Shorter names make a designed-for case more likely to be seen. That is not worth
uniquifying logic. Nothing to do.

The rest of the review stands — the `kept_worktrees` readers and the
`restore.rs:57` gate are the two that matter.

EmilSkaaning:
Thanks — all of it addressed. Summary of where each point landed.

**Fixed in this round**

- **Compile break in `tests/create_e2e.rs`.** Two full `SpawnRequest` literals were
  missing `existing_worktree`, not one: `recipe-probe` and `env-probe` (the second
  arrived with #1047). Both fixed in 6c53bb4c.
- **`SCHEMA_VERSION` doc comment** stopped at v41 while the constant was 42 — v42
  clause added in 18a277d5.
- **Asymmetric arms in `recreate_worktrees`.** The borrowed arm pushed
  unconditionally while the thurbox arm guarded on `branch_exists`, so a restore
  could hand a session a cwd that was gone. Now guards on `worktree_path.is_dir()`,
  with tests for both outcomes (06f40bc4).
- **PR description** now states the scope: opening is TUI-only, since
  `thurbox-cli session create` has no `--worktree-path`.

**Already on the branch when you reviewed**

The v42 renumber (`storage/sessions.rs:46`, `sync/state.rs:103`,
`docs/FEATURES.md:2004`), all three `kept_worktrees` readers, the restore gate
(`force_delete_was_lossy`), the "three reads" → "four" comments, and the `create()`
path validation.

**Found in a follow-up pass**

- `migrate_from_v41_backfills_worktree_provenance_as_thurbox_made` — the migration
  had no test, breaking a pattern thirteen others keep. Backfilling `0` instead of
  `1` would make force-delete skip a worktree it owns.
- `a_borrowed_worktree_stays_borrowed_across_the_wire` — the JSON round-trip was
  covered only for the absent-key default, so a writer that stopped emitting
  `created_by_thurbox` would read back as "ours" on a peer and hand a borrowed
  worktree to `git worktree remove --force`, with every test still green.
- `.claude/skills/thurbox-cli/SKILL.md` still described the old unconditional
  teardown and restore refusal; `website/docs/features.html` still said "Closing a
  session automatically removes the worktree".

Both new tests were checked by mutation, not just observed green.

**Not changed:** the session-name collision point, which you withdrew.

LeTuR:
Second pass, on `dcee5ff6` rebased onto `5f7032a6`. Everything from the first
review is addressed, and the two additions I did not ask for — the v42 backfill
test and the provenance-on-the-wire test — are the two I would have asked for
next. `cargo nextest run --all` on the merged tree: 2313 pass, 0 fail.

One finding, and it is half mine.

## The `is_dir` guard does not prevent the cwd it cites

`06f40bc4` skips a borrowed worktree whose directory is gone, and its comment
gives the reason as: "restoring the row would hand the session a cwd that is not
there". The guard does not stop that. `recreate_worktrees`' return value is read
exactly once, for a count:

- `restore.rs:160` — `let recovered = recreate_worktrees(&deleted.worktrees);`
- `restore.rs:173` — `worktrees_recovered: recovered.len()`

Nothing is written back. `restore_session` (`storage/sessions.rs:178`) clears
`deleted_at` and `force_deleted` on the sessions row and touches nothing else, so
the stored `cwd` is untouched. `respawn` at `restore.rs:162` then reloads the row
and `build_restart_plan` anchors on `session.cwd` (`restart.rs:91`), which
`resolve_launch_cwd` returns unchecked for a single-worktree session
(`spawn.rs:842-844`). The pane still opens at the missing path.

What makes that reachable is my `force_delete_was_lossy`: before it, a
force-deleted session was refused outright unless `--best-effort`, so nobody got
here. Now a session whose worktrees were all borrowed is restored without a
prompt — including when the directory is gone.

```mermaid
flowchart TD
    A["force-delete: borrowed worktree kept"] --> B["user deletes their own checkout"]
    B --> C["restore"]
    C --> D["force_delete_was_lossy = false<br/>→ allowed, no --best-effort"]
    D --> E["recreate_worktrees skips it<br/>→ worktrees_recovered: 0"]
    E --> F["session row restored, cwd untouched"]
    F --> G["respawn opens a pane at the missing path"]
```

The guard is still worth keeping — it stops the report claiming a worktree came
back when it did not, which is real. But it is a reporting fix, not the
prevention its comment describes.

Cleanest resolution is to make the gate answer the question the restore actually
faces. `force_delete_was_lossy` asks "could the teardown have destroyed
anything"; the missing case is "can this restore deliver what it promises". A
borrowed worktree that is gone fails the second even though it passes the first,
and refusing it with its own message — the worktree is no longer on disk, rather
than the misleading "uncommitted and untracked changes are gone" — is both
honest and consistent with how the thurbox arm behaves. Alternatively keep the
skip and reword the comment to claim only the count. Either is fine; the current
pairing is the one combination that reads as handled and is not.

## nit

- `restore.rs:173` reports `worktrees_recovered: 0` for this case and
  `respawn_error` will likely carry the tmux failure, so it does surface — just
  not as the reason.

## Confirmed fixed

`kept_worktrees` reaches all three readers including the ADR-24 delegated parse;
no `schema v41` references remain for the provenance column; `create`'s doc block
is back on `create` with `session_name` carrying its own; the SCHEMA_VERSION
header has its v42 clause. The website correction in `dcee5ff6` catches a claim
that was already false before this PR ("Closing a session automatically removes
the worktree"), which is the #1044 lesson applied without being asked.

Note the suite passes here only with `THURBOX_SOCKET_FOR` unset — that is #1048,
unrelated to this branch and already green.

EmilSkaaning:
Pushed as `3e7e3d03`. Your finding stands and is fixed; following it to the other
callers turned up two more instances of the same mistake.

## The `is_dir` guard

You were right that it only ever changed a count, and right about why that is
reachable — `force_delete_was_lossy` is what stopped refusing this case. Taking
the first of your two resolutions.

`session_ops::restore::restore_refusal` replaces the boolean gate and answers
both questions: *could the teardown have destroyed anything* (unchanged) and
*can this restore deliver what it promises*. The second refuses a borrowed
worktree that is no longer on disk, naming the path rather than the uncommitted
work that was never touched. It is not conditioned on `force_deleted` — a soft
delete followed by the user removing their own checkout lands in exactly the
same place, and `--best-effort` remains the way to say yes to either.

The skip stays, since keeping `worktrees_recovered` honest is real; its comment
now claims the count and points at `restore_refusal` for the prevention.

## Two more readers of the raw flag

Both are the same error the guard's comment was: treating `force_deleted` as the
answer to "is this restorable".

- **`cli/sessions.rs:1001`** re-derived the refusal itself, so `thurbox-cli
  session restore` still refused a borrowed-only force-delete that the pipeline
  has allowed since `force_delete_was_lossy` landed. The two interfaces
  disagreed, which that function's own doc says must not happen. It now calls
  `restore_refusal` and adds only the `--best-effort` sentence.
- **`ui/plugins/80_restore.lua`** keyed its confirmation on `partial`
  (= `row.force_deleted`), wrong in both directions: it confirmed a
  borrowed-only force-delete with "uncommitted and untracked changes were lost
  on delete", which is false, and once the new refusal existed a soft-deleted
  row with a missing borrowed worktree would have been refused with no way to
  override from the TUI. `DeletedRow` now publishes `restore_refusal` — the
  kernel's own sentence, nil when the restore would simply run — and the pane
  asks iff it is set, showing that reason instead of one written in Lua.
  Computing it stats each borrowed worktree, so it sits in the refresh that runs
  on `data_version` movement, not in a render path.

`partial` is kept alongside it: it says how the row was deleted, which is a
different fact from whether the restore is refusable, and still drives the tag
and the muted styling. The new test asserts exactly that split — a row still
tagged `force-deleted` that restores with no question.

Also `thurbox.yml:291`, a fifth "three parameterised reads" comment the earlier
sweep missed — and one that had drifted above `thurbox.settings`, which is not
one of them.

## Checks

2276 pass. Both branches of the new gate were mutation-checked rather than
observed green: `!best_effort && deleted.force_deleted` fails the
missing-worktree test, and dropping `force_delete_was_lossy` fails two others.
clippy, `RUSTDOCFLAGS="-D warnings" cargo doc`, rumdl, selene, stylua and
lua-language-server are clean. `docs/FEATURES.md` and the `thurbox-cli` skill
now name `restore_refusal` as the single decision behind all three surfaces.

## nit

Agreed on `worktrees_recovered: 0` — with the refusal in front of it that path
is now only reachable via `--best-effort`, where the count is the report the
caller asked for rather than a missed reason.

LeTuR:
Third pass, on `3e7e3d03` (base `5f7032a6`). `cargo nextest run --all`: 2321 pass.
fmt, clippy `-D warnings`, `cargo doc`, `cargo deny`, architecture rules, rumdl,
selene, stylua and `lua-language-server --check` all clean. One finding, which
I pushed the fix for myself (`ceb4e2b3`) rather than bounce back to you — it is
two lines and it falls out of the same reasoning as your `create` guard.

## `restore_refusal` asks the local disk about a path on another machine

`restore_refusal`'s second question — is the borrowed worktree still there —
is a bare `worktree_path.is_dir()`, and it runs for every deleted row, remote
sessions included. A session on an `ssh:` host has its checkout on the host, so
that path was never meant to exist locally: the stat fails, and the row reads as
unrestorable over a directory that is in fact still sitting there. The TUI shows
the same wrong sentence, since `SnapshotStore` computes the refusal the same way.

Reachable exactly as far as the feature goes: `list_worktrees_on` takes a host and
`want_worktrees` carries one, so opening an existing worktree on a remote host
works — and then that session can never be restored without `--best-effort`.

You already drew this line once, in `create`:

```rust
if host.is_none() && !worktree.is_dir() {
    return Err(format!("not a worktree: {}", worktree.display()));
}
```

The fix is the same line in the same spirit — `restore_refusal` takes the
backend type and skips the disk check when it is remote. Nothing is lost by
skipping it: `restore_session_headless` delegates a remote restore to the host's
own `session restore`, which reaches this function again with the paths on the
filesystem they belong to. The lossy force-delete branch is untouched, since that
one is not a disk question.

Test first, and it fails on the pre-fix code with the exact wrong sentence:

```text
left: Some("'borrowed' opened the worktree at /repo/.worktrees/mine, and it is
           no longer on disk; …")
right: None
```

`docs/FEATURES.md` and the `thurbox-cli` skill both now say the second check is a
local session's only.

## Everything else

The rest holds up. Provenance is threaded end to end — column, wire, `WorktreeInfo`
round trip, teardown, restore — and the v42 backfill defaults the safe way.
`parse_worktree_list` drops what it should and keeps `locked`. The `↳` rows cost
one `git worktree list` per highlighted repo with no fetch, which is the right
trade for a keypress. Naming after the directory rather than the branch is the
call I'd have made.

Nothing else blocking. CI is running; merging once it is green.

Reviewed as LeTuR's agent.

https://claude.ai/code/session_01P2VZUmdf9qRbF9q5yzbSDZ

LeTuR:
Thank you for your contribution @EmilSkaaning !

EmilSkaaning:
You are very welcome, @LeTuR .
Thank you for making this tool! It is awesome

PR #1049, opened 2026-09-01
feat(ui): confirm session delete when an open PR would be walked away from

TMSCH:
## Summary
- Add `open_pr` to the per-session git snapshot, populated best-effort via `gh pr view` on the same background thread that computes `worktree_stats`.
- Publish it to Lua as `session.git.open_pr` (nil when unknown or PR is not OPEN).
- Extend the sessions plugin's `at_risk` check so force-deleting a session whose branch has an open PR asks first and names the number.

## Why
Force-deleting a session removes the worktree and the branch's local checkout. If a PR was still open on that branch, the delete would silently walk away from the reviewer's comments, the CI runs, and the branch the PR is anchored to. The plugin already asks first for uncommitted changes and unpushed commits — an open PR belongs in the same list.

## Test plan
- [x] `cargo nextest run --test session_list` — 14 pass, including a new `force_deleting_a_session_with_an_open_pr_asks_and_names_the_number`.
- [x] `cargo check --all` — clean.
- [x] `cargo clippy --all-features --lib --bins -- -D warnings` — clean.
- [x] `cargo test --test architecture_rules` — pass.
- [x] Manual: with `gh` missing/unauthenticated, `open_pr` stays `None` and no question fires (silent absence, per plugin comment).

### DIFF
--- a/src/git/mod.rs
+++ b/src/git/mod.rs
@@ -27,32 +27,33 @@ mod plugin;
 mod remote;
 mod worktree;
 
-/// Look up an open PR for the worktree's current branch via `gh pr view`.
+/// Whether the worktree's HEAD is already reachable from `origin/HEAD` — i.e.
+/// the branch has been merged into origin's default branch.
 ///
-/// Returns `Some(pr_number)` only when the PR is OPEN — merged and closed PRs
-/// report `None`, and any `gh` failure (missing binary, no auth, no remote, no
-/// PR) is a `None` too. Best-effort by design: the session-delete confirmation
-/// treats absence as "no reason to ask".
-pub fn open_pr_number(worktree_path: &Path) -> Option<u64> {
-    let output = Command::new("gh")
-        .args(["pr", "view", "--json", "state,number"])
+/// `Some(true)` = merged, `Some(false)` = ahead of / diverged from the default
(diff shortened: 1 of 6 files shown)


LeTuR:
Superseded by #1054, which takes the forge-agnostic route this PR was reaching for — but by a different mechanism, because the one here does not survive a squash merge.

`git merge-base --is-ancestor HEAD origin/HEAD` returns exit 1 on a squash-merged branch: the merge rewrites the work into one new commit, so none of the branch's own commits are ever reachable from the default branch. Verified in a scratch repo. Since this project squash-merges exclusively, the check would have reported `Some(false)` for every branch that had in fact landed — and the PR added that as a *new* reason to prompt, on top of the false `ahead > 0` reason that was already firing on every finished session.

#1054 keeps the `--is-ancestor` fast path for merge commits and fast-forwards, then compares **patches** for the squash case: square the branch off into one throwaway commit on its merge base and ask `git cherry` whether the default branch already carries that patch. Local refs only, same as here — no `gh`, no network — so GitHub, GitLab, Bitbucket and a bare repo behind an SSH remote all answer alike. It removes the false prompt rather than adding to it.

https://claude.ai/code/session_01EUntdytakxnECjDQ8rVwk3

PR #1107, opened 2026-09-11
fix(core): announce a program pane's exit and give each window the remain-on-exit it needs

spscream:
A program started by a plugin — the editor, for one — could exit without thurbox noticing. Quit vim with `:q` and the pane stayed a corpse: a second click or Enter on a file did nothing, because the pane still looked alive.

Three things were wrong, and each one alone is enough to lose the exit.

**`remain-on-exit` was a session option that isn't one.** It sat in `SESSION_OPTS`, but tmux resolves `set-option -t <session> remain-on-exit` down to the session's *current window* (measured on tmux 3.2a). `apply_session_config` runs on every `ensure_ready`, so which windows carried it was an accident of timing — and the two roles want opposite answers. An agent's window should keep its corpse, so the error it died with stays readable and a listing can still report `#{pane_dead}`. A plugin's program is read from its output *stream*, and tmux announces a pane's death only by closing its window — so a kept window is a death that is never announced. It is now stated per window **in the same tmux command list that creates the window** — `on` for `tb-`, `off` for `tbs-`/`tbp-` — so neither role's correctness waits on a second message: a command that exits instantly used to take its window, and the server with it when that window was the last one (measured, tmux 3.2a: five such windows lost every time when the option was sent separately, kept every time when it was chained). `window-size manual` moved to a `WINDOW_OPTS` list applied with `set-option -w -g`, which also gives every window the `off` default.

**Only one spelling of the close notification was parsed.** A program exiting on its own emits `%unlinked-window-close`, not `%window-close`. Both are parsed now and turned into a program-exit event the kernel and plugins can see.

**A corpse nobody could see is a corpse nobody could clear.** `find_window` silently skipped dead panes, so the guard meant to clear one never fired — the probe showed two windows of the same name, `@1 dead=1` and `@2 dead=0`. It is replaced by `window_panes`, which reports every window of the name together with its `#{pane_dead}`; `find_program_window` kills the corpses and returns the live pane.

### Testing

`tests/window_remain_on_exit.rs` asserts both roles end to end on a real tmux, through both real spawn paths, because what is under test is the wiring — a helper returning the right string proves nothing about which windows are told. `tests/program_pane_corpse.rs` and `tests/program_pane_exit.rs` cover the other two halves. Every test was checked with its fix stubbed out: the corpse test gives `left: 2, right: 1`, the option test `left: "<unset>", right: "on"`. Suite is green on `cargo nextest run` apart from four tests that fail identically on unpatched `main` here (they read the machine's real `hosts.toml` and expect `wsl:ubuntu`).

### Rejected

- *Verifying pane liveness inside `start_program`* — it is documented as being asked on every frame by plugins, so that would put a tmux round trip on a per-frame path. The cheap check lives only on the re-adoption branch.
- *Keeping `remain-on-exit` for program windows* — the corpse says nothing, and keeping it is exactly what makes the exit unannounceable. Worth revisiting only if a program's last screen must survive its exit, and then the corpse needs pairing with another liveness signal.
- *Querying tmux for live panes from the control reader thread on `%window-close`* — the response comes back through that same thread, so waiting on it deadlocks. The pane→window mapping is recorded at `register_pane` time instead.
- *Inverting the birth value* (`on` server-wide, `off` per window) — it swaps which role loses the race rather than removing it, and hands the loss to the role that cannot absorb it: tmux announces a pane's death only by closing its window, so a program pane born `on` and dying inside the round trip is a death that is never announced. Chaining the option into the creating command list removes the race for both.

### Not in this PR

Two additive plugin APIs — `keys` (typing into a program a plugin already started) and `on_context` (the right mouse button) — were in this branch and have been taken out of it. They want their own review and neither is touched by `program.exited`; each comes back as its own PR with the API described in the body.


spscream:
Both review findings are addressed in `ee27214`, each with a test that fails with the fix stubbed out.

**Adopted windows were taken as found.** `start_program` finds a program window by its deterministic name, and that window was made by an *earlier* interface — possibly one that set `remain-on-exit` session-wide and landed it on whichever window was current. Left as found, an editor window carrying `on` from that era is a pane whose exit can never be announced, and it comes back on the first restart after the upgrade — the one moment this change exists to fix. `adopt` now calls `normalise_remain_on_exit`, which reads the window name and sets the value that name's role wants.

Covered by `tests/window_remain_on_exit.rs::adopting_a_program_window_normalises_what_it_finds`: start a program pane, force `remain-on-exit on` on its window, drop the `Terminals`, build a second one over the same tmux, ask for the same `ProgramKey` (the adoption path), assert the window reads `off`. Without the fix: `left: "on", right: "off"`.

**A restart could swallow `program.exited`.** The loop applies commands before it derives events, so a plugin asking for its program on the frame after it died restarts the pane, and `program_liveness` then hands the deriver a live pane under the same key — no transition, no event. The ending is now recorded where the slot is overwritten (`Terminals::replaced_program_exits`) and drained in `coordinator::events` **before** the memo is replaced, under the same rule the walk uses ("only a pane we had seen running"), so the restarted pane's own entry cannot make it fire twice.

Covered by the new `tests/program_restart_exit.rs`, driven through the real restart path on a real pane: run a program that ends on its own, wait for `has_exited`, ask again, assert the ending was kept. Without the fix: `left: [], right: [ProgramKey { plugin: "plugins/90_files.lua", name: "editor_opts" }]`.

Full suite: 2626 run, 4 pre-existing environment-dependent failures on this machine (they fail identically on a clean `main` worktree — they read the machine's real host list and expect a `thurbox-cli` binary).

spscream:
I had the branch reviewed independently as well, which found four more. All fixed in `38dd88a`.

**A program that died before the loop looked was never announced.** The loop applies commands, talks to tmux and serves its workers before it derives, so a program with a bad argument — or a missing binary inside a wrapper, or one that prints usage and exits — is already gone by the time the transition is derived, and was never in the memo of panes seen running. The gate that dropped it is there to ignore corpses adopted from a *previous* run of the interface, which is a different thing: that program stopped while nothing was watching, and announcing it at boot would tell a pane its editor had just closed. Freshly spawned panes are now recorded as they are spawned (`Terminals::take_started_programs`), so the gate keeps its meaning and stops swallowing real endings.

**The derivation had no test at all.** It was a block inside a method on `App`, which nothing in the suite constructs — deleting `enqueue_event` from it left every test green. It is a free function over plain data now (`program_endings`: three readings in, endings out, memo updated in place), with seven tests: the ordinary transition, ending once rather than once per frame, the adopted corpse that must stay silent, the ending a restart replaced, a pane that died twice in one iteration (two programs really did end, so two events), and the addressing that two plugins both calling their pane `editor` depend on. Two of them fail with the relevant line stubbed out.

**Adoption paid two subprocess round trips per pane.** `normalise_remain_on_exit` sat at the head of the generic `adopt`, which carries every agent pane on every ssh host — two RTT each, in front of the parallel prefetch ADR-P9 exists to keep that path fast. It has moved to `find_program_window`, where the window was just looked up *by name*: the role is known without asking, so it is one round trip, local, and only for the program windows it was written for. `SessionBackend::set_pane_retention` is the seam, defaulting to a no-op for a backend with no window options.

**A window could still be born keeping its corpse.** The per-window setting is applied after `new-window`, so a user with `set -g remain-on-exit on` in `~/.tmux.conf` — read on thurbox's socket too — gets a program window that keeps a pane which died inside that round trip, and that death is never announced. `off` is now the birth value in `WINDOW_OPTS`; the one role that wants a corpse asks for it.

Two smaller things in the same commit: the tests in this area reported a broken spawn as a skipped environment and passed, which is the exact failure `program_pane_exit.rs`'s own header records having been caught by once — tmux being installed is checked at the top, and everything after it is now a failure. And `register_pane` losing the pane-to-window mapping says so in the log instead of silently costing that pane its death notice.

Suite: 2573 run, the same four pre-existing environment-dependent failures on this machine.

spscream:
The P1 on `38dd88a` is right, and it is the kind that only shows up in the
sequence: **an ending announced from the live walk was announced again by the
restart that followed it.** Confirmed before touching anything —

```
FAIL coordinator::events::tests::an_ending_already_announced_is_not_announced_again_by_the_restart
     one process died and ended twice: ["editor"]
```

The deriver took two unordered readings — panes spawned since the last look,
and endings whose slot a restart had already overwritten — seeded the memo from
the spawns and drained the replacements against it afterwards. A restart on a
*later* frame therefore put the surface back into the memo, and the replacement
it carried vouched for a death that had already been reported. A plugin that
restarts its program on `program.exited` restarted twice for one process. The
second half of the finding is the same false claim from the other side: the key
was recorded as started before the spawn could fail.

The two sets cannot be fixed by reordering them, because the answer depends on
the order: a death is news only if its occupant was started after the last drain
or was known running before it. Both readings are now **one ordered log** —
`Started` / `Replaced`, in the order they happened — walked in sequence. A
`Replaced` consumes the memo entry it reports, so nothing after it can report
the same death; a `Started` vouches only for the pane it spawned, and is
recorded after the spawn succeeds.

The documented behaviour is unchanged where it was deliberate: a pane that died,
was restarted and died again inside one iteration is still two endings, because
two programs really did end. `docs/PLUGINS.md` now states the other half too —
one death, one event.

Covered at both levels, and the order is what is asserted:
`tests/program_restart_exit.rs` expects the log to be exactly
`[Replaced, Started]` on a real tmux pane, since the order is the fix. Sorting
the log to put spawns first — the old reading — fails both that test and the
unit one.

One more thing this branch needed: it did not pass `cargo fmt --check`. Only the
review bot runs on a fork PR, so nothing said so; `efc96ab` fixes it.

src/kernel/terminal/mod.rs
```diff
@@ -518,12 +535,13 @@ impl Terminals {
...
-    /// narrower signal than it looks. Control mode carries every pane on a
-    /// backend down one connection, so this fires when the *connection* goes —
-    /// a host or ssh dropping, a local tmux server dying — and **not** when a
-    /// single pane is killed. A restart therefore cannot be caught here, however
-    /// tempting it looks: it says so itself, through [`Terminals::forget`].
+    /// `has_exited` is set when a session's output **stream** ends. Control mode
+    /// carries every pane on a backend down one connection, so it fires when the
+    /// *connection* goes — a host or ssh dropping, a local tmux server dying —
+    /// and, since the kernel started reading tmux's window-close notifications,
+    /// when a single pane's window closes as well. A restart is therefore caught
```
LeTuR, line 542: `has_exited()` now means two different things on a remote backend, and the branch below still reads it as one.

`.agents/skills/thurbox-session-status/SKILL.md:70` states the old invariant outright: "with `remain-on-exit=on` a clean agent exit keeps its pane alive (no reader EOF), so a remote reader hitting EOF means the host/SSH connection dropped". `drop_lost_panes` acts on exactly that inference — for a remote row it does `self.ready.borrow_mut().clear()` and records `"host unreachable"` against the session. After this change a single `%unlinked-window-close` on an ssh host produces the same EOF, so one window closing gets diagnosed as the host going away, and it clears the readied-backend cache for *every* backend, not just that host's.

How reachable is it in practice? `SessionBackend::kill` unregisters the pane before it kills, and `forget` beats the notification on the restart path, so thurbox's own teardown is fine. What is left is a window killed from outside — someone in tmux on the host, or a `tmux kill-window` from a script.

Is the remote arm still right, or does it now need to tell a window close from a connection drop before it declares the host unreachable? I do not have a strong view on which way to resolve it and I would rather you picked.

The doc is not optional either way: CLAUDE.md's Rule is that a change invalidating a documented decision updates it in the same PR, and the skills are covered by that — the file says so itself. That skill line is wrong as of this branch.

src/agent/tmux.rs
```diff
@@ -410,6 +410,33 @@ impl WindowRole {
...
+/// Should a window of this name keep its pane's frame after the pane dies?
+///
+/// Yes for an agent (`tb-`) and no for everything else, and the difference is
+/// **how each one's death is noticed**:
+///
+/// - A session's liveness is read from a *listing* (`#{pane_dead}`, see
+///   [`WindowIndex`]), which a kept window answers truthfully. Keeping it is the
+///   point: the agent's last screen — the error it printed — stays attachable,
+///   and `live_agent_window` still refuses to call the corpse an agent.
+/// - A shell (`tbs-`) and a plugin's program (`tbp-`) are read from their pane's
```
LeTuR, line 432: `tbs-` gets the same `off`, but nothing reads a companion shell's exit — so the frozen pane that cannot be reopened, which is the failure this PR closes for programs, stays open for shells.

`Session::has_exited()` (`src/agent/backend.rs:561`) loads `self.wired.exited`, the agent's pane only. `ShellPane`'s own flag has no reader anywhere in the tree — I grepped every `has_exited()` call site. And `ensure_shell_pane` (`src/agent/backend.rs:1309`) returns early on `self.shell_pane.is_some()` without consulting it. So: type `exit` in a `Ctrl+T` shell, the window now reliably closes, the reader gets its EOF, the flag gets set, and the user still sees a dead grid that swallows every key, with `Ctrl+T` refusing to give them a new one.

Not a regression — the same thing happened before, just by accident rather than by design. But this line commits to "a shell's liveness is read from its output stream", and nothing reads it. Either drop the shell pane when its reader ends, or say here that a shell's exit is deliberately not acted on yet and why. I will take the note; I would rather not have the claim stand unqualified.

src/kernel/command/mod.rs
```diff
@@ -203,6 +203,20 @@ pub enum Command {
...
+        /// closing it, and the one that lets a long-lived program be *told*
+        /// something rather than replaced. Restarting an editor to open a
+        /// second file is the case that asked for it: the process is the
+        /// expensive part, and `start_program` is idempotent, so without this
+        /// the only way to change what it shows was to close it first.
+        ///
+        /// Sent WITH a program, it means "type at it, or start it if it is not
+        /// running" — the coordinator picks, because whether the pane is alive
+        /// is not something a plugin can read. See `apply_program`.
+        keys: Option<String>,
```
LeTuR: `keys` is a new plugin API and it is in neither the title nor the PR body — I found it only in the diff. `on_context` is the same shape of addition and gets one line in the body. Neither is a program pane noticing its own exit, and neither has a caller in `ui/`, so nothing in the shipped interface exercises either one.

Split them out. Concretely, what I will do: review and merge the window-lifecycle fix on its own as soon as the two points above are settled — that is the bug fix and I do not want it waiting on anything else. `keys` and `on_context` each come back as their own PR with a body describing what they are for.

For `keys` specifically I want to see it next to a bundled pane that uses it. "Send keys, or start the program if the pane is not there" is a contract with a silent failure mode on both sides — the keys are dropped when the pane turns out to be missing and the fallback start puts the program somewhere the keys would not have, and a plugin cannot tell which happened. That is worth its own review, not a paragraph in this one.

src/agent/tmux.rs
```diff
@@ -1402,6 +1437,34 @@ impl TmuxBackend {
...
+        // ability to notice its own death, and neither of which used to leave a
+        // trace: the round trip failing (a reconnecting control mode, a
+        // multiplexer without `display-message`), and the pane dying inside the
+        // round trip — the window close then arrives before there is anything
+        // to match it against. The first is logged here. The second is narrow
+        // and deliberately not paid for: closing it means asking again after
+        // registering, which is a second round trip on every pane attached, to
+        // catch a program that died in the millisecond it took to ask once.
+        let window_id =
+            match self.ctrl_command(&format!("display-message -t {pane_id} -p '#{{window_id}}'")) {
```
LeTuR, line 1591: nit: this round trip looks buyable for free, and it is on the path the third commit already went out of its way to protect.

`spawn` asks tmux for the pane id at line 1742 with `-P -F '#{pane_id}'`; `-F '#{pane_id} #{window_id}'` costs nothing more. On the adopt side `window_panes` and `DISCOVER_FORMAT` are both `list-windows -F` and could carry `#{window_id}` the same way. Either would leave `register_pane` with the mapping already in hand and no `display-message` at all — which is the same argument you make for taking `normalise_remain_on_exit` off the generic `adopt`, and it matters for the same reason: every agent pane on every ssh host comes through here.

Same line of thinking, one more: now that `WINDOW_OPTS` (line 2848) carries `remain-on-exit off` globally, the per-window write at line 1766 is setting `off` on a window that was already born with it. Only the agent's `on` still needs saying.

Not blocking — it is one serialized control-mode round trip, not an ssh connection — but if you are already touching the spawn format string it is close to free.

src/agent/tmux.rs
```diff
@@ -1744,7 +1841,21 @@ impl SessionBackend for TmuxBackend {
         self.capture_history_seed(backend_id)
     }
 
-    fn find_window(&self, window_name: &str) -> Result<Option<String>> {
+    fn set_pane_retention(&self, backend_id: &str, keep: bool) -> Result<()> {
```
LeTuR, line 1960: nit, security hygiene rather than a bug: this is the only new mux-reaching method without the `is_valid_pane_id` guard every neighbour carries (`capture_history` 1838, `stamp_window` 1916, `adopt` 1789, `kill`, `resize`).

Not exploitable today, and I checked rather than assumed: the one caller passes an id `window_panes` already validated, and `tmux_run` goes through `TmuxTransport::prefixed`, which POSIX-quotes every token, so there is nothing to inject even over ssh. But this is a public trait method, and tmux resolves `-t <target>` as a name as readily as an id, so a future caller passing something that is not a pane id would quietly set `remain-on-exit` on whatever window that name picked out. Two lines, and it matches the file.

LeTuR (changes requested):
Read the whole diff and the surrounding code for each point below. Writing as LeTuR's agent; the merge call is the operator's.

The core reading is right, and I checked the premise rather than taking it: on tmux 3.7c here, a program exiting on its own in a `new-window` of the attached session emits `%unlinked-window-close @1` and never `%window-close`, so parsing only the first spelling really does mean parsing none of them. That holds five minor versions past the 3.2a you measured on. The `SESSION_OPTS` diagnosis matches too.

Since this is a fork and only the review bot runs on it, I ran the gate: `cargo fmt --all --check` clean, `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo nextest run --all` **2574 passed, 0 skipped**. The four failures your body mentions on `main` do not reproduce in a worktree with no `hosts.toml`, so nothing here is hiding behind them.

Security pass, since this touches command construction: **no findings.** `set_pane_retention` and `window_panes` go through `tmux_run` -> `TmuxTransport::prefixed`, which POSIX-quotes every token, so the ssh and WSL arms are covered. The new `ctrl_command` interpolations (`display-message`, `set-window-option`) take pane ids that `is_valid_pane_id` has already reduced to `%<digits>`. `keys` reaches the pane through the keystroke writer, which is `send-keys -H` hex on tmux and the quoted `-l` encoder on psmux, so arbitrary bytes from a plugin cannot become tmux tokens. Window names stay behind the fixed `tb-`/`tbs-`/`tbp-` prefixes and `sanitize_window_name`, so a plugin's pane name cannot make its window look like an agent's. One hygiene nit inline. On process lifetime: `off` for `tbs-`/`tbp-` strictly reduces what outlives its owner, and `on` for `tb-` is the behaviour the repo already intended — but it now reaches every agent window rather than whichever one was current, so dead agent panes will accumulate where they previously did not. Nothing reaps them short of deleting the row. That is the pre-existing design, not a regression, and I am not asking you to change it here.

Requesting changes on three things, two of which are process rather than code:

1. `has_exited()` now means two things on a remote backend and one documented consumer still reads it as one. A skill states the old invariant outright and has to change in this PR.
2. `keys` and `on_context` are two new plugin APIs riding in a window-lifecycle fix. Split them out — detail inline.
3. The companion shell gets the new `off` and nothing reads its exit, so half the fix lands. I will take either a fix or a written note that it is deferred.

Everything else I would merge as it stands. The ordered-transition-log rework in the last commit is the right shape for the problem and its tests pin the thing that was actually wrong.

Not commenting on: naming, formatting, import order, or the test files' skip-versus-panic policy, which you already tightened in the third commit.

spscream:
Thank you for running the gate — and for checking the premise on 3.7c rather
than taking the 3.2a measurement on trust.

This round is the red CI plus the two points that do not need a decision. The
one that does is answered below, not in code.

**CI was one root cause, not five.** `cog.toml` allows `api, cli, ui, git, core,
docs, deps, config, mcp`, and this branch used `program` — so every commit and
the PR title failed, and Nextest / Windows were *cancelled* rather than failing.
All five subjects and the title are `fix(core)` now, which is the scope the
neighbouring tmux work on `main` uses (`fix(core): stamp tmux windows with an
owner id`). Force-pushed, so the hashes moved.

**The retention guard** — `is_valid_pane_id` on `set_pane_retention`, with the
reason you gave: `-t` resolves a name as readily as an id, and this is a public
trait method whose next caller is not the one you and I read.

**The companion shell** — taking the note, and it is on `keeps_dead_pane` where
the claim is made. `off` makes the shell's death *reportable* and nothing
reports it: `Session::has_exited` reads the agent's pane alone, `ShellPane`'s
flag has no reader, `ensure_shell_pane` returns early on a filled slot. Dropping
the shell pane when its reader ends is the fix, and it is a different change —
it decides what happens to a pane someone is looking at, where this one only
decides whether tmux announces anything.

**`has_exited()` on a remote backend — deferred, deliberately, and I would like
your read before I pick.** You asked me to choose between "the remote arm is
still right" and "it has to tell a window close from a connection drop", and
that choice changes what `drop_lost_panes` does to every session on a host. I
would rather resolve it in its own PR, with the skill line, than guess at it
inside this one. If you would rather the skill line not stand wrong for that
long, say so and I will correct the wording here and leave the behaviour
question to the follow-up.

**`keys` and `on_context` — agreed, they should not be here.** Same reasoning
for deferring the split to a separate push: both are additive and pulling them
out rewrites this branch a second time. Which order do you want — this branch
stripped first and the two APIs re-proposed after, or all three at once?

Everything else on this branch is unchanged; the local gate here is `fmt`,
`clippy --all-targets` and 2574 tests, with the four failures my body mentions
reproducing on `main` in this environment and nowhere else.

spscream:
Both branches are rebased on `main` (through #1110) and force-pushed; #1107 was
seven commits behind, which is also why `current_wsl_distro` was missing from it
locally. Gate re-run on the rebase: `fmt`, `clippy --all-targets`, 2630 tests,
same four environment failures as `main` here.

The review bot's `keys` finding on this push is right and is fixed in
`cad3e7f`: `ui/lib/thurbox.d.lua` told plugin authors `keys` "starts nothing",
while `apply_program` starts the program from `repo` when the send finds no live
pane — and drops the keys it was given. `docs/PLUGINS.md` had it right; the file
an editor reads over the author's shoulder did not. Both branches are named
there now, the dropped keys included.

That is the same silent failure mode you raised when asking for `keys` to come
out of this PR, arrived at from the other end. It does not change the split —
the contract still deserves its own review next to a bundled pane that uses it.

The workflow runs are sitting at `action_required` after each force-push, so
Actions has not re-run since your approval this morning.

src/kernel/terminal/mod.rs
spscream: Not changed here, and deliberately so: on a remote backend `has_exited()` would have to mean one of the two things, and whichever it means changes `drop_lost_panes` for every session on that host. You wrote that you would rather I picked, so the pick is a follow-up with the `.agents/skills/thurbox-session-status/SKILL.md:70` line in the same diff, rather than a semantics change smuggled into a PR about program panes — the reasoning is in https://github.com/Thurbeen/thurbox/pull/1107#issuecomment-5645730137.

If you would rather the skill line stopped saying the wrong thing now, say so and I will correct the wording in this branch and leave the behaviour alone.

src/agent/tmux.rs
spscream: Confirmed, and only the note is in this branch (`bb21ce5`, on `keeps_dead_pane`): `Session::has_exited` reads the agent's pane alone, `ShellPane`'s flag has no reader, and `ensure_shell_pane` returns early on a filled slot — so the shell's `off` makes its death reportable while nothing reports it.

The fix decides what happens to a pane someone is looking at; this PR only decides whether tmux announces anything. Held back for its own change, as you offered to take a written note instead.

src/kernel/command/mod.rs
spscream: Agreed on both counts, and the split is the follow-up. What is fixed here is the half that was actively wrong rather than merely undocumented: `ui/lib/thurbox.d.lua` claimed `keys` starts a pane, which it does not — on its own it starts nothing and says so when the pane holds nothing live; sent *with* `repo` a pane with nothing running is started from it and the keys are dropped (`cad3e7f`).

`keys` and `on_context` come out as their own PRs with the API documented in the body, unless you would rather I drop them from this branch first.

src/agent/tmux.rs
spscream: Not acted on, and I would rather it were its own change than a rider here. `-F '#{pane_id} #{window_id}'` is free at the format string, but it changes what `spawn` parses, what `window_panes` parses and `DISCOVER_FORMAT` — the adopt path for every pane on every host — and that wants a test of its own that fails when the second field goes missing. Happy to do it in this PR if you would rather have it now.

On the second half: the per-window `off` at 1766 is redundant whenever the `-w -g` write landed, which is the common case. I left it because the global write is best-effort and the two cover different failures — a server whose options thurbox did not get to set still gets the right default from the per-window write. Say the word and it goes, leaving only the agent's `on`.

src/agent/tmux.rs
spscream: Fixed in `bb21ce5`:

```rust
if !control_mode::is_valid_pane_id(backend_id) {
    bail!("refusing to set remain-on-exit on invalid pane id: {backend_id:?}");
}
```

Same guard, same place in the call as the neighbouring mux-reaching methods.

LeTuR: Addressed in `bb21ce5f`, and it is the guard I asked for — same `bail!` shape as `capture_history` immediately above it, so the file reads consistently now.

One thing I checked rather than assumed, since it is the branch where a locally-correct change goes wrong: the guard sits *before* the psmux early return. On psmux an invalid id used to return `Ok(())` and set nothing; now it returns `Err` and sets nothing. Same effect, and the sole caller (`find_program_window`) already logs at debug and carries on, so there is no new failure path on the Windows side. Windows (clippy) and Windows (tests) both pass on this head. Settled, nothing further from me.

src/agent/tmux.rs
LeTuR: Taking the note — this is what I asked for, and putting it on `keeps_dead_pane` where the claim is made is the right place for it.

Agreed on the split too: dropping the shell pane when its reader ends decides what happens to a pane someone is looking at, which is a different question from whether tmux announces anything. Settled. I am not asking for the shell fix in this PR and I will not ask for it in review of the next one either — it wants its own change.

src/agent/tmux.rs
LeTuR: Still open, still a nit, and it does not gate anything — no reply needed on it.

Leaving it as: worth doing whenever something next touches that format string, and fine to drop if you disagree. I would rather you spent the re-push on the two points that actually block.

src/kernel/terminal/mod.rs
LeTuR: Answering, since you asked me to pick: **correct the wording here, leave the behaviour to a follow-up.** That is the split you offered and it is the right one.

The reason for not letting it ride: the sentence is not vague, it is an inference — "a remote reader hitting EOF means the host/SSH connection dropped" — and this branch makes that false today. The next person to read `drop_lost_panes` and reason from it reaches the wrong conclusion, and that is a worse failure than the open behaviour question, because it looks settled. Correcting the wording is not deciding the behaviour, so it does not pre-empt the follow-up.

What I want in the skill: that `has_exited()` now means the stream ended, which on a remote backend is either the connection going or that pane's window closing; that `drop_lost_panes` still treats it as host loss and clears the readied-backend cache for every backend; and that telling the two apart is open. A sentence or two — do not re-litigate it in the file.

The behaviour decision is mine, not yours. I will carry it in the follow-up rather than have you guess at it, because it changes what happens to every session on a host and that is a maintainer call. You are not blocked on it beyond the wording.

src/kernel/command/mod.rs
LeTuR: Answering the ordering question: **strip this branch first, re-propose the two APIs after.**

This is a bug fix with a symptom a user can hit, and four commits of convergence behind it; it should not queue behind the review of two additive APIs. Removing them is also the smaller, more mechanical change — I checked that the split is clean: `program.exited` does not touch either of them, and `docs/PLUGINS.md`'s exit section stands on its own. What comes out is `keys` (`kernel/command`, `coordinator/commands`, `kernel/host/api`, `thurbox.d.lua`, its two `docs/PLUGINS.md` sections) and `on_context` (`coordinator/mouse`, `kernel/host/mod`, `thurbox.d.lua`, `docs/PLUGINS.md`, `tests/mouse_context.rs`).

On `cad3e7f1`: the corrected annotation is right and I would have asked for it. It does not change the split, as you say — if anything it is the argument for it, since the contract needed a correction before a single caller existed.

src/agent/tmux.rs
```diff
@@ -2763,15 +2895,41 @@ fn history_seed_bytes(mut raw: Vec<u8>) -> Vec<u8> {
...
     ("window-size", "manual"),
+    // The default a window is BORN with, so the one role that wants a corpse
+    // asks for it and nothing else inherits one. Set here as well as per window
+    // at spawn because the two cover different moments: the user's
+    // `~/.tmux.conf` is read on thurbox's socket too, and `set -g
+    // remain-on-exit on` there would have every window born keeping its corpse
+    // — including a program that dies in the round trip between `new-window`
+    // and the option being set on it, whose death would then never be
+    // announced.
+    ("remain-on-exit", "off"),
```
LeTuR, line 2988: This birth value is, I believe, why Nextest is red — and it is the mirror image of the race the same commit set out to close.

`spawn_window` (the headless one-shot path the failing tests use) does, in order: `ensure_session_configured` -> `apply_session_config` -> this list, so the server is now globally `remain-on-exit off`; then `new-window`; then, at line ~2020, `set-window-option -t <pane> remain-on-exit on` for a `tb-` window. The window is therefore *born* `off`, and the write that makes it an agent window arrives afterwards. A command that has already exited by then — a missing agent binary, which #1104 has just made a supported state rather than an error — takes its window with it, and if that window is the last one on the server, tmux exits. The next spawn then reports exactly what CI reports: `tmux new-window exited exit status: 1 for window tb-flow: server exited unexpectedly` (the string is `spawn_window`'s own, which is how I know it is this path and not control mode).

Both failures fit that shape: each is the *second* install in its test, after a first one whose `flow` agent does not exist on the runner.

I could not reproduce it through the suite — 2630/2630 pass here, as they do for you — so I tested the mechanism directly:

```
# birth value off (this line), agent window with a missing binary, last window on the server
tmux -L x new-session -d -s thurbox -n tb-flow -x 80 -y 24 'not-a-real-binary'
tmux -L x set-option -w -g remain-on-exit off
sleep 1.5
tmux -L x new-window -t thurbox -n tb-flow2 'sleep 300'
#   -> no server running

# birth value on (what the branch replaced)
#   -> tb-flow dead=1, server up, next new-window returns %2
```

I am not certain this is what CI hit — I cannot run the runner's environment — so please confirm before acting on it. What makes me think it is: it is the only mechanism in the diff that produces that exact string; the failure is on `tb-` windows, whose role is the one that wants `on`; and Nextest is green on #1104, #1110 and on your own #1108, all on this same base.

On the fix, a suggestion rather than a requirement, since you have thought about this ordering more than I have: making `on` the birth value and writing `off` per window for `tbs-`/`tbp-` inverts which role loses the race, and the program side now degrades gracefully where the agent side does not — a program that dies in the round trip keeps a corpse, and `find_program_window` already kills corpses on the next lookup. That machinery exists because of this branch. The agent side has no such recovery: the server is simply gone.

Whichever way you take it, the thing I want out of it is that neither role's correctness depends on winning a round trip against a process that can exit instantly.

LeTuR (changes requested):
Re-reviewed `cad3e7f1`. Answers to both of your questions are in the threads they belong to, so this is the ledger plus one new thing CI found.

**Where the five points stand.**

| | point | status |
|---|---|---|
| 1 | `has_exited()` on a remote backend + the stale skill line | **deferred by agreement** — I picked: correct the wording here, behaviour to a follow-up, and that decision is mine to carry |
| 2 | split `keys` and `on_context` out | **agreed, not yet done** — I picked the order: strip this branch first |
| 3 | companion shell's `off` buys nothing today | **addressed** (`bb21ce5f`) — the note, on `keeps_dead_pane` where the claim is made |
| 4 | nit: `register_pane`'s extra round trip | untouched, still a nit, does not gate |
| 5 | nit: `set_pane_retention` missing the id guard | **addressed** (`bb21ce5f`) |

Nothing dropped and nothing silently re-argued. Points 3 and 5 are done and I am not coming back to them.

**What I checked on the delta.** `git range-diff` says the four original commits are byte-identical apart from the `fix(program)` -> `fix(core)` subject, so the only new code is `bb21ce5f` (+17 in `tmux.rs`) and `cad3e7f1` (a `thurbox.d.lua` annotation). I also checked the rebase for a semantic conflict rather than trusting a clean apply: #1104 is the only new commit on `main` touching files this PR touches, and it changes `ControlMode::start`'s error mapping and refactors `WindowIndex::locate` — neither of which this branch goes near. No conflict.

**Security re-review on the changed code: no new findings, and point 5 closes one.** The guard strictly narrows a mux-reaching path; I confirmed the psmux ordering is behaviour-neutral (in the thread). `cad3e7f1` is a `---@field` line in a `---@meta` file nothing loads into the VM, so it has no runtime surface at all. Command construction, names as an injection surface, exit status as a forgeable signal, and the remote/WSL/psmux branches are unchanged from the head I already cleared, and nothing in the delta reopens them.

**The new thing, and the reason this is still CHANGES_REQUESTED rather than an approval of a branch that only needs a doc line: Nextest is red, and I think this branch is why.** Two tests fail — `session_ops::extensions::tests::uninstall_reverses_install` and `::reinstall_tears_down_then_installs_fresh` — both with `tmux new-window exited exit status: 1 for window tb-flow: server exited unexpectedly`, both on `TRY 3`, so not a one-off. Detail and a reproduction are in the new thread on `WINDOW_OPTS`. Short version: `remain-on-exit off` as the birth value means an agent window whose command exits immediately closes, and when it is the last window on the server the server goes with it — the per-window `on` is written after `new-window` returns and cannot save a process that has already exited. That is the same round-trip race your third commit closed for program panes, running the other way for agent panes.

I could not reproduce it in the suite — 2630/2630 pass here on `cad3e7f1`, as they do for you — so I went at the mechanism directly and it reproduces cleanly against tmux. Windows (tests) and Windows (clippy) are green, so this is the Linux one-shot spawn path specifically.

**So what is left, in the order I would do it:**

1. The Nextest regression. This is the one that actually blocks — everything else on the list is a line of prose or a deletion.
2. The skill wording, per my answer in that thread.
3. Strip `keys` and `on_context`.

After those three I expect to approve; there is nothing else outstanding from my side, and I have no new objections to the code itself. The window-lifecycle work is the part I want and it has not changed since I said so.

Still not merging and not approving on your behalf — fork PRs are the operator's call, and that has not changed.

spscream:
Confirmed, and it is that line — with one correction to the fix, because the
inversion trades the failure rather than removing it.

**Your mechanism reproduces.** The job log has the two failures you predicted,
`session_ops::extensions::tests::uninstall_reverses_install` and
`reinstall_tears_down_then_installs_fresh`, both with `spawn_window`'s own
string, both through three retries — not a flake. And the mechanism reproduces
directly here on tmux 3.2a, deterministically, which your first attempt could not
get through the suite: five windows created with `sh -c 'exit 7'` and
`remain-on-exit` sent as a second call were gone every time (`no such window`);
with the option chained into the *same command list* as `new-window`, the corpse
was kept every time. A command list runs to completion before the server returns
to its event loop, so there is no moment in it for a pane to be reaped.

**Why not the inversion.** It swaps which role loses, and the role it hands the
loss to is the one that cannot absorb it: tmux announces a pane's death only by
closing its window, so a `tbs-`/`tbp-` window born `on` and dying inside the
round trip is a death that is never announced — the frozen pane you described in
the other thread, and the bug this branch exists to fix. `find_program_window`
does reap a corpse, but only when someone next looks that program up.

**So neither role waits on a round trip.** `retention_suffix` puts the option in
the command list that creates the window, on both spawn paths (`bb21ce5` →
`9d57f6a`). Both answers are stated rather than only the one that differs from
the default: the server-wide write is best-effort, and a program window that
inherited `on` from a user's `~/.tmux.conf` because that write failed is exactly
the unannounced death again — chained, saying it costs no round trip, which also
answers your `1766` nit.

The control-mode path leaves the target unsaid: `new-window` without `-d` makes
the window it created current, so the bare form is that window — including when
an older window shares its name, where `-t <name>` resolves to the lowest index
instead (measured). The headless path passes `-d` and has to name its window, so
that collision stays exactly where it was.

`tests/window_remain_on_exit.rs` gains the case: an agent whose command exits at
once still leaves its window. It fails deterministically with the option sent
separately (`left: None`), which is the CI failure reproduced in the suite.

**The split — done, your ordering.** `4a2cacb` takes both APIs out: `keys` from
the `program` command, its Lua reader and its two `docs/PLUGINS.md` sections;
`on_context` from the mouse path, the host, the type declarations and
`tests/mouse_context.rs`. Four files and one test file leave the diff entirely
(`coordinator/commands.rs`, `coordinator/mouse.rs`, `kernel/host/api.rs`,
`kernel/command/mod.rs`, `tests/mouse_context.rs`), 414 lines net. What stayed
from that neighbourhood is `dispatch_event`'s filtering by `event.only` — that
is how `program.exited` reaches only the plugin that started the pane, not part
of either API. The PR body is updated too: the `on_context` paragraph and its
rejected bullet are gone, and there is a "Not in this PR" note saying where the
two are going.

It is a removal commit rather than a rewritten history, so the branch still
shows them arriving and leaving. Say the word if you would rather have the
history folded.

**The skill line — corrected here, behaviour left to you** (`337816f`). It now
says `has_exited()` means the pane's stream ended, which on a remote backend is
either the connection going or that pane's window closing; that `drop_lost_panes`
still reads it as host loss and clears the readied-backend cache for every
backend on the host; and that telling the two apart is open. No re-litigating in
the file.

**The `-F '#{pane_id} #{window_id}'` nit** is untouched, as you asked.

Local gate on `4a2cacb`: `fmt`, `clippy --all-targets` clean, 2628 tests run,
2624 passed — the four failures are the environment ones that fail identically
on `main` here.

src/agent/tmux.rs
spscream: Confirmed — and the mechanism, not just the shape. The job log has your two tests with `spawn_window`'s own string, three retries each. And it reproduces here deterministically, which the suite would not do: five windows created with `sh -c 'exit 7'` and `remain-on-exit` sent as a second call were gone every time (`no such window`); chained into the same command list as `new-window`, the corpse was kept every time (tmux 3.2a).

I did not take the inversion, for one reason: it swaps which role loses rather than removing the loss, and the role it hands it to cannot absorb it. tmux announces a pane's death only by closing its window, so a `tbs-`/`tbp-` window born `on` that dies inside the round trip is a death that is never announced — the frozen pane from your other thread, and the bug this branch is about. `find_program_window` does reap a corpse, but only when someone next looks that program up.

What `9d57f6a` does instead is put the option in the command list that creates the window, on both spawn paths. A command list runs to completion before the server returns to its event loop, so there is no moment in it for a pane to be reaped — neither role waits on a round trip, which is what you asked for either way.

Two details worth having: the control-mode path leaves the target unsaid, because `new-window` without `-d` makes the window it created current and the bare form is exactly that window — `-t <name>` would resolve to the lowest index when an older window shares the name (measured). The headless path passes `-d` and has to name its window, so that collision stays where it was. And both answers are stated, not only `on`: the server-wide write is best-effort, and a program window that inherited `on` from a user's `~/.tmux.conf` because that write failed is the unannounced death again. Chained, saying it is the same message — which also answers your 1766 nit.

`tests/window_remain_on_exit.rs` gains the case; it fails with `left: None` when the option is sent separately.

src/kernel/command/mod.rs
spscream: Done in `4a2cacb`, in that order. Both APIs are out: `keys` from the `program` command, its Lua reader and its two `docs/PLUGINS.md` sections; `on_context` from the mouse path, the host, the type declarations and `tests/mouse_context.rs`. Your file list was exact — `coordinator/commands.rs`, `coordinator/mouse.rs`, `kernel/host/api.rs`, `kernel/command/mod.rs` and `tests/mouse_context.rs` leave the diff entirely, 414 lines net.

One thing stayed from that neighbourhood on purpose: `dispatch_event` filtering by `event.only`, which is how `program.exited` reaches only the plugin that started the pane rather than every subscriber. It sits next to the removed code but belongs to this PR.

The PR body is updated the same way, with a "Not in this PR" note. It is a removal commit rather than a rewritten history — say so if you would rather the history were folded.

src/kernel/terminal/mod.rs
spscream: Corrected here in `337816f`, behaviour left to your follow-up, exactly as you split it.

The line now says `has_exited()` means the pane's stream ended; that on a remote backend this is either the host/SSH connection going or that pane's own window closing, `remain-on-exit=on` keeping a clean agent exit from ending the stream; that `drop_lost_panes` still reads it as host loss and clears the readied-backend cache for every backend on that host; and that telling the two apart is open. Two sentences and a clause, no re-litigating in the file.

src/agent/tmux.rs
```diff
@@ -2881,6 +3068,17 @@ pub fn spawn_window(
...
 
+    // Chained into the same command list as the creation, not sent after it —
+    // `retention_suffix` has the measurement. This path passes `-d`, so the new
+    // window is not current and the option has to name it; a same-named window
+    // from another session with the same name would take the write instead
+    // (tmux resolves such a target to the lowest index), which leaves that rare
+    // case exactly where it was before.
+    if !cfg!(windows) && keeps_dead_pane(&window_name) {
+        tmux.args([";", "set-window-option", "-t", &window_name]);
+        tmux.args(["remain-on-exit", "on"]);
```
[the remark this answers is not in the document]
spscream: Right, and it was the one residual I flagged in the commit rather than closed. Fixed in `090d638`.

The headless path passes `-d`, so the control-mode path's bare target — the window `new-window` just made current — is not available there. It now creates the window at the end of the list instead (`-a -t {end}` appends after the last), so `{end}` inside the same command list is exactly the new window, whatever it is called, and the older namesake keeps its own value.

Measured on tmux 3.2a and pinned by a test that spawns two windows of the same name with the second exiting at once: it fails when the target goes back to the name.

LeTuR:
**A third machine does not clear this branch — and `main` on that same machine is 2606/2606 green.** The two named tests fail at `cad3e7f1`, and they still fail at `4a2cacb3`, after `9d57f6a`. I ran this as a second environment for the CI failure, on a box that is neither the reviewer's nor the author's: Debian 13, **tmux 3.5a**, rustc 1.97.1, 4 cores.

## The mechanism reproduces, exactly as described

tmux 3.5a, own socket, verbatim. The dying window is the server's only window:

```
$ tmux -L tbxrace set-option -w -g remain-on-exit off
$ tmux -L tbxrace new-window -d -t tb: -n tb-flow -P -F '#{pane_id}' /bin/false
-> rc=0 out=%1
$ tmux -L tbxrace set-window-option -t %1 remain-on-exit on   # the second call
no server running on /tmp/tmux-1000/tbxrace
-> rc=1
$ tmux -L tbxrace new-window -d -t tb: -n tb-flow-2 -P -F '#{pane_id}' 'sleep 600'
-> rc=1 out=no server running on /tmp/tmux-1000/tbxrace
```

With the birth value `on`, same script: `set-window-option` returns 0, the window is there as `@1 tb-flow dead=1`, and the second `new-window` returns `%2`.

With a session that still has its own shell window, the birth value `off` costs the option but not the server — `no such window: %1`, then `@0 zsh dead=0`, and the next `new-window` returns `%2`.

Timing sweep, 726 attempts firing `new-window` across the moment the last window dies: 573 `no server running`, 127 fine, **16 `server exited unexpectedly`**, 10 `no current target`. That last-but-one string is the one in the CI log, so the race does produce it on 3.5a and not only the `no server running` form.

## `9d57f6a`'s premise holds on 3.5a — the chained form is atomic

Chaining `set-window-option` into the same command list as `new-window`, 5 runs each, birth value `off`, command `/bin/false`:

```
named target  (-d, headless):  run 1..5  rc=0  windows=[zsh dead=0  tb-flow dead=1]
bare target   (control mode):  run 1..5  rc=0  windows=[zsh dead=0  tb-flow dead=1]
```

The corpse is kept every time, in both spellings. So the fix does what it says at the tmux level on this version too.

## The suite, at three SHAs on that machine

`cargo nextest run --all`, one run each:

| SHA | | two named tests | full suite |
|---|---|---|---|
| `f7f2024f` | `main` | **pass** | **2606 run, 2606 passed, 0 failed** |
| `cad3e7f1` | reviewed head | **fail** | not run to completion |
| `4a2cacb3` | current head, after `9d57f6a` | **fail** | 2628 run, 2610 passed, **18 failed** |

Both failures carry the strings from the job log, at both SHAs:

```
tmux new-window exited exit status: 1 for window tb-flow: server exited unexpectedly
tmux new-window exited exit status: 1 for window tb-flow: no server running on /tmp/tmux-1000/thurbox-dev
```

They are not a flake and not cross-test interference: each fails alone under `-E 'test(...)'`, fails under `--test-threads=1`, and failed on every one of the nine runs I did of each (three rounds of: the pair in parallel, the pair serialized, each one on its own). The failing call is the *reinstall* — `install_extension` at `tests.rs:1048`, the one right after `uninstall_extension` tore the session down.

The other 16 at `4a2cacb3` are the rest of the tmux surface: `window_remain_on_exit` (all three, including this branch's own `an_agent_that_dies_at_once_still_leaves_its_window`), `program_pane_exit`, `program_pane_corpse`, `program_restart_exit`, `spawn_command_resolution`, `concurrent_respawn`, `session_lifetime` (two), `cli_stdout_contract`, `tui_e2e` (five).

**None of them fail on `main` here.** That is the part worth acting on: the four environment-dependent failures you and the reviewer both see do not exist on this box, so "it is the machine" does not cover these.

Sampling the socket at 10 ms through a single run of `uninstall_reverses_install`:

```
main       @0 zsh dead=0
           @0 zsh dead=0   @1 tb-flow dead=0     <- agent window up and alive
           @0 zsh dead=0                          <- torn down, base window and server survive  -> PASS

4a2cacb3   @0 zsh dead=0                          <- tb-flow never sampled
           no server running                      -> FAIL
```

The run is 0.10 s, so that is only a handful of samples and I would not call it proof of where the window goes — but the contrast with `main`, same host, same tmux, same shell, is what it is.

## The inversion, tried

`("remain-on-exit", "off")` → `"on"` in `WINDOW_OPTS`, on `cad3e7f1`:

- The two named tests: **still fail.** It does not fix them.
- `window_remain_on_exit`, `program_pane_exit`, `program_pane_corpse`: 4 run, 4 failed — three of them by hitting a 20 s timeout, which is the never-announced death you argued it would trade for. (Caveat: those same tests already fail at `4a2cacb3` unmodified on this box, so I cannot hand the inversion sole credit for them. Note also that `an_agent_that_dies_at_once_still_leaves_its_window` does not exist at `cad3e7f1` — it arrived with `9d57f6a`.)

So the inversion is not the fix here, which agrees with your reasoning against it from the other side.

## The asymmetry

It is really there in the code, but it is not load-bearing the way the review states.

- `start_program` → `find_program_window` does reap: for every pane of that window name, `dead` gets `backend.kill`, and a live one gets `set_pane_retention(false)`. `spawn_window` has no equivalent.
- But that reaper is about corpses left by an *earlier* interface, and it runs on the next lookup. It does nothing for a pane that dies inside the `new-window` round trip, and nothing at all once the server is gone — there is no one left to reap for.

The asymmetry that mattered at `cad3e7f1` is narrower: the birth value is `off`, and `off` is already what the program and shell roles want. Only the agent role needed the second call to say something different, so only the agent role depended on a round trip. That is why inverting only moves which role is exposed, and why `retention_suffix` stating both answers in the creating command list is the right shape.

## What I did not establish

I did not find the root cause of the 18. The mechanism is real and `9d57f6a` closes it, and yet these tests fail here and pass on `main` — so something else on this branch is also reaching them, and I stopped short of bisecting the nine commits to find it. Happy to run that, or anything else, on this host if it is useful.

Everything above is one run per cell unless a count is given; the box is idle and long-lived, so any of it can be repeated.

— LeTuR's agent

src/agent/tmux.rs
LeTuR: Still red at `090d6385`: CI run 34702467686, same two extension tests, same `server exited unexpectedly`, all three tries. The green run at this head is `PR Title`, not `CI`.

The chaining itself is right — I checked it: the retention lands, `{end}` beats an older namesake, and `-t <name>` does resolve to the lowest index. It just is not what these two tests are dying on.

One place to look, as a question rather than a claim: `ensure_session_configured` creates the session's own bootstrap window and no retention is ever stated for it, so on this branch it is born with the server-wide `off` — where on `main` it was the one window `set-option -t <session> remain-on-exit on` actually landed on. If that shell exits, the server goes with it. I could not reproduce locally (2629/2630 here), so please confirm before acting on it.

— LeTuR's agent

LeTuR:
Not approving yet, and the reason is only the one: CI's Nextest still fails at `090d6385` — run 34702467686, the same two extension tests, the same `server exited unexpectedly`, all three tries. The green run at this head is `PR Title`, not `CI`. Detail and one place to look are in the `WINDOW_OPTS` thread.

Everything else is done and I have resolved those four threads: the skill wording, the shell note, the retention guard, and the split — `on_context` is gone entirely and `keys` with it, while `program.exited` is intact. I verified the new targeting rather than taking it on trust: the chained retention lands on the new window, `{end}` beats an older namesake, and `-t <name>` does resolve to the lowest index.

The round-trip nit stays open as a marker, not an ask — `new-window` still asks only for `#{pane_id}` though the format string moved. It does not block, and I am happy for it to be dropped.

— LeTuR's agent

spscream:
**The bootstrap window is not it — CI was dying on `window-size`, and that one reproduces here deterministically.** I built tmux 3.5a locally to test the hypothesis and got the failure on the first try, which is why the hypothesis could be answered rather than argued: on tmux 3.2a this branch cannot fail, on 3.5a it cannot pass.

## Your hypothesis, measured — and it is not the cause

It is true that the branch dropped a protection `main` had. `set-option -t <session> remain-on-exit on` lands on the session's current window, which right after `new-session` is the bootstrap window, so on `main` that window kept its corpse and on this branch it does not:

```
main-like:   window 0 remain-on-exit = on     (global stays off)
branch-like: window 0 remain-on-exit = <unset, inherits global off>

bootstrap shell exits ->  main-like: [zsh dead=1]      next new-window -> %1
                          branch-like: no server running    next new-window -> rc=1
```

But nothing kills that shell in the failing runs. Watching the socket through the failing test, `bash:%0` is alive and the server is gone a moment later — and arming `remain-on-exit on` on window 0 before the test does **not** save it. The server is not losing its last window; it is dying.

## What it is: `window-size manual` as a server-wide default is a NULL dereference

`spawn_window` asks for the size of a window that does not exist yet — `default_window_size(sc->tc, s, NULL, …)` — and with the manual type `clients_calculate_size` does

```c
} else if (type == WINDOW_SIZE_MANUAL) {
        *sx = w->manual_sx;      /* w is NULL */
```

so the server dies inside `new-window`. The server's own `-vv` log from the failing test stops exactly there, between `spawn_window: name=tb-flow` and `spawn_pane: shell=`.

Minimal, raw tmux, nothing of thurbox in it:

```
tmux 3.5a  manual + plain `new-window -d`   -> rc=1  server exited unexpectedly
tmux 3.5a  auto   + plain `new-window -d`   -> rc=0  %0 %1
tmux 3.5a  session-scoped manual (main)     -> rc=0  bash:80x24 x:80x24
tmux 3.2a  manual + plain `new-window -d`   -> rc=0  %0 %1
```

`default-size` does not help; neither does the target spelling. The guard `w != NULL` exists only on tmux master — 3.3, 3.4, 3.5a and 3.6 all read the NULL. 3.2a predates the manual size entirely, which is why your box, my box and the reviewer's all said the mechanism was fine.

This is the branch's own regression, and a plain one: on `main` both `remain-on-exit on` and `window-size manual` sat in `SESSION_OPTS`, where the session target quietly landed them on the session's current window and on no other window ever. Moving `window-size` to a server-wide `-w -g` write — correct in spirit, a window option has no session scope — made **every** spawn fatal on tmux ≥ 3.3, headless and control mode alike (a control client is ignored for sizing, so it takes the same branch).

## The fix, `f01fc59`

`window-size manual` is stated **per window, in the same command list that creates the window** — the same place and for the same reason as the retention, which is now one `birth_options` list rather than a retention-only suffix. `WINDOW_OPTS` keeps `remain-on-exit off` server-wide (that one is harmless and still wanted for windows thurbox did not create) and must not name `window-size` again.

The intent survives: a window thurbox creates still sizes itself rather than follow the smallest attached client — strictly more coverage than `main`, where exactly one window (the bootstrap) ever had it.

## Verification

| | tmux 3.2a (this box) | tmux 3.5a (built here) |
|---|---|---|
| `main` (`f7f2024`) | pass | **pass** (both named tests) |
| branch `090d638` | pass | **fail**, `server exited unexpectedly`, both tests |
| branch `f01fc59` | pass | **pass**, both tests |

Full suite on 3.5a at `f01fc59`: **2631 run, 2630 passed**, the one failure being `resolve_host_accepts_the_backend_name_the_interface_carries`, which fails identically on `main` here. On 3.2a: 2631 run, 2628 passed, the three environment failures that also fail on `main`.

Two guards, both of which fail when the server-wide write is put back (checked, not assumed): `WINDOW_OPTS` may not name `window-size` (unit), and a real spawned window carries `manual` while the server does not (`tests/window_remain_on_exit.rs`, version independent — it asserts the configuration, which fails the same way on a 3.2a machine that can never see the crash).

## Left alone, deliberately

The bootstrap window still inherits `off` on this branch where `main` gave it `on` by accident. That is a real behavioural difference and it is measured above, but it is not what CI hit, and giving thurbox's own session window a corpse is a decision rather than a fix — say the word if you want it stated explicitly, and I will chain it into `new-session` the same way.

The retention chaining stays as it was: its own measurement still holds (five windows lost unchained, five corpses kept chained), it is just not what the job log was reporting. The comment that claimed it was is corrected in the same commit.

LeTuR (approved):
CI is green on f01fc593 and the Nextest blockers from the last two passes are fixed: retention now rides in the same command list as new-window, so the corpse is kept and the server stays up.
One regression goes with it and is tracked as a follow-up: window-size is not supported at the tmux 3.2 floor, and it now sits in the command list that creates every non-psmux window.

— LeTuR's agent

src/agent/tmux.rs
LeTuR: Not taken in this PR. It is a marker rather than an ask and does not block the merge.

— LeTuR's agent

src/agent/tmux.rs
LeTuR: The red CI this thread was about is green now: All Checks and CI both pass on f01fc593, with retention in the same command list as new-window.

— LeTuR's agent

src/agent/tmux.rs
```diff
@@ -420,6 +420,109 @@ impl WindowRole {
...
+        // without checking — a NULL dereference that takes the whole server
+        // down. Measured, tmux 3.5a: with `set-option -w -g window-size
+        // manual`, *every* `new-window` on a server with no attached client
+        // answered `server exited unexpectedly`; with the same option said per
+        // window it answers with a pane id. Unguarded in every release that has
+        // the option (3.3 … 3.6; guarded only on tmux master), and 3.2a — the
+        // supported floor — predates it. Stating it after the window exists is
+        // what `main` did by accident, where a session-scoped write landed on
+        // the session's current window and on no other.
+        ("window-size", "manual"),
```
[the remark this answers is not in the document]
LeTuR: Confirmed, and deferred to a follow-up fix rather than held here: gate window-size on the tmux version that introduced it, and stop an unsupported option from orphaning the window.

— LeTuR's agent

PR #1108, opened 2026-09-11
fix(core): give Ctrl+V to the agent when the clipboard holds an image

spscream:
Pasting a screenshot into an agent pane did nothing. `Ctrl+V` is a global binding that always claims the press, and thurbox's clipboard transport carries text — so the image was dropped and the press never reached the CLI in the pane, which is the one thing here that can read an image.

Claude Code fetches it itself on seeing `Ctrl+V` — its trigger is the key event (`if(e.key!=="v")return false; if(e.ctrl&&!e.meta) return platform!=="windows"`), and it then shells out to `xclip`/`wl-paste`, or to PowerShell under WSL. So the fix is to stop swallowing the chord: `kernel.paste` declines with `Some(false)` when there is no text to paste — the same answer `kernel.copy` already gives with no selection — and the press falls through to the focused terminal.

**Inside WSL that decision cannot be made locally.** WSLg bridges the clipboard's *text* only. Copy an image in Windows and the X clipboard is not updated at all: it keeps handing out the last text copied. Measured here — `Clipboard::get_text()` returned an IP address copied minutes earlier while Windows held a 1594x535 PNG — so the paste inserted something *stale* rather than nothing, which is the worse half of the bug. arboard cannot see the image either: thurbox builds it with `default-features = false`, the build without `get_image`.

So `clipboard::ImageProbe` asks `powershell.exe` whether the Windows clipboard holds an image. That costs ~0.42 s (three runs: 0.42/0.41/0.43), which is far too long to hold the event loop for and is paid on *every* paste, not just image ones — so it is another instance of the worker pattern: the press is claimed, the interface keeps drawing, and the answer decides who gets it. Nothing is asked off WSL, where the local clipboard is the one being copied into.

A forwarded chord is dropped while a modal is open, for the reason a pasted text is: it must not leak into the terminal behind the overlay.

### Testing

`only_a_wsl_distro_asks_windows` pins the gate — it is the whole cost control, since a gate that answered "yes" everywhere would add a third of a second to every `Ctrl+V` on every platform.

`the_windows_probe_agrees_with_powershell` runs the real round trip against the real Windows clipboard, because the wiring is what breaks: the argument list, `-Sta` (without it `Add-Type` throws and every answer becomes "no image"), the `PATH`/interop fallback, and reading the answer off the exit code. It skips where it cannot discriminate — off WSL, with no PowerShell, or with no image on the clipboard — and deliberately never writes to the clipboard: a suite that destroys what you copied is worse than one that skips. Both tests were checked with the fix stubbed out and fail.

### Rejected

- *A second chord for "give the paste to the agent"* — cheap and exact, but it leaves the ordinary `Ctrl+V` after copying an image still pasting stale text, which is the half that corrupts a prompt silently.
- *Detecting the image with arboard* — the feature is not in this build, and the X clipboard it would read does not carry the Windows image anyway.
- *Asking PowerShell on the loop* — 0.42 s of frozen interface per paste.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -144,13 +144,30 @@ last text copied — a paste there inserts something stale rather than nothing,
 which is worse. `clipboard::ImageProbe` asks `powershell.exe`
 (`Clipboard::ContainsImage`) instead. That costs ~0.42 s measured, which is why
 it is the seventh instance of the worker pattern rather than a call on the loop:
-the press is claimed immediately and acted on when the answer lands, by which
-time focus is re-resolved. Off WSL nothing is asked — the local clipboard is the
-one being copied into, and arboard's answer is the whole truth.
+the press is claimed immediately and acted on when the answer lands. Off WSL
+nothing is asked — the local clipboard is the one being copied into, and
+arboard's answer is the whole truth.
+
+Three bounds make that latency safe. The **target** is the session the press was
+aimed at, taken at the press: 0.42 s is long enough to change panes, and a paste
+landing in a pane nobody aimed it at corrupts what is being typed there. **One
+question runs at a time** — key auto-repeat makes presses tens of times faster
+than the answer, and a `powershell.exe` per repeat is a held key bringing the
+machine down — with the presses that arrive while it is out kept (at most eight)
+and asked about separately, because an answer may only classify presses that
+predate it. And the child has a **five-second deadline**, after which it is
+killed and read as "no image": WSL interop can wedge outright, and the question
+after this one waits on its answer.
 
 **Rejected**: *a second chord for "give the paste to the agent"* — cheap and
 exact, but it leaves the ordinary `Ctrl+V` after copying an image still pasting
 stale text, which is the half of the bug that corrupts a prompt silently.
+*Resolving the focus when the answer lands* — it reads as "paste where the
+person is looking", but the intent belongs to the press, and a paste that
+arrives in a pane it was not aimed at is the same corruption seen from the other
+side. *Spending one answer on every press waiting behind it* — the clipboard can
+change while the question is out, and a press classified by what preceded it is
+exactly the stale paste this section is about.
 *Detecting the image with arboard* — thurbox builds it with
 `default-features = false`, the build without `get_image`, and the X clipboard
 it would read does not carry the Windows image anyway.
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -129,6 +129,32 @@ are handled in one place.
   crossterm's internal byte representation doesn't match xterm
   sequences. Modifier keys, in particular, would break.
 
+### An image on the clipboard is the agent's paste, not ours
+
+thurbox's clipboard transport carries text (`clipboard::copy`/`paste`, and the
+OSC 52 leg by construction). An image therefore cannot be pasted here at all —
+but the CLI in the pane can fetch one itself when it sees the paste chord, so
+`Ctrl+V` is **handed to it** rather than swallowed. `kernel.paste` declines the
+chord, the same `Some(false)` answer `kernel.copy` uses when there is no
+selection, and the press falls through to the focused terminal.
+
+Inside WSL the decision cannot be made locally: WSLg bridges the clipboard's
+*text* only, so while Windows holds an image the X clipboard still hands out the
+last text copied — a paste there inserts something stale rather than nothing,
+which is worse. `clipboard::ImageProbe` asks `powershell.exe`
+(`Clipboard::ContainsImage`) instead. That costs ~0.42 s measured, which is why
+it is the seventh instance of the worker pattern rather than a call on the loop:
+the press is claimed immediately and acted on when the answer lands, by which
+time focus is re-resolved. Off WSL nothing is asked — the local clipboard is the
+one being copied into, and arboard's answer is the whole truth.
+
+**Rejected**: *a second chord for "give the paste to the agent"* — cheap and
+exact, but it leaves the ordinary `Ctrl+V` after copying an image still pasting
+stale text, which is the half of the bug that corrupts a prompt silently.
+*Detecting the image with arboard* — thurbox builds it with
+`default-features = false`, the build without `get_image`, and the X clipboard
+it would read does not carry the Windows image anyway.
+
 ### A paste on Windows arrives as keys, not as a paste
 
 `Event::Paste` is unix-only. crossterm's Windows source reads console
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -156,8 +156,10 @@ than the answer, and a `powershell.exe` per repeat is a held key bringing the
 machine down — with the presses that arrive while it is out kept (at most eight)
 and asked about separately, because an answer may only classify presses that
 predate it. And the child has a **five-second deadline**, after which it is
-killed and read as "no image": WSL interop can wedge outright, and the question
-after this one waits on its answer.
+killed and the press handed to the agent unanswered: WSL interop can wedge
+outright, and the question after this one waits on its answer. A probe that was
+killed is never read as "no image" — that reading is exactly the stale paste
+this path exists to stop.
 
 **Rejected**: *a second chord for "give the paste to the agent"* — cheap and
 exact, but it leaves the ordinary `Ctrl+V` after copying an image still pasting
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -2524,7 +2524,10 @@ confined to the active pane bounds.
   newlines, the multi-line task description keeps them). While **any**
   modal is open the paste is swallowed so it can never leak into the
   terminal in the pane behind the overlay; otherwise it pastes into
-  the active PTY.
+  the active PTY. The rule is about the **press**: a press made while
+  an overlay is up never had a pane to name. An answer owed to a press
+  made *before* the overlay went up is still delivered where that press
+  was aimed — see "Pasting images".
 - **`Ctrl+V` with an image on the clipboard**: handed to the agent in
   the pane instead of pasted — see "Pasting images" below.
 - **`Ctrl+Shift+V`** (your terminal's paste): the way to paste when
@@ -2644,7 +2647,16 @@ answered by a question of their own, and a question that has not come back in
 five seconds is abandoned — an unanswerable question is handed to the agent, not
 read as "no image", because the text it would paste instead is the stale one.
 Nothing is asked off WSL, where the local clipboard is the one being copied
-into.
+into, and nothing is asked on a distro where no `powershell.exe` could be found
+twice running — there is nobody to answer, and every paste would pay the wait.
+
+**A late answer still goes where the press was aimed.** No question is asked
+while a modal or a float owns typed input, so an overlay stops pastes starting
+underneath it. An overlay that goes up *after* the press, while the question is
+out, does not redirect or cancel it: the pane was named when the key was pressed,
+and a paste that vanished because someone opened a float for a third of a second
+would be the silent loss this path exists to end. Because that delivery can land
+out of sight, it says so on screen.
 
 **On macOS the byte is synthesised.** `Cmd+V` is the paste binding there, and a
 `Cmd` chord has no pty encoding at all — handing it on by declining it would
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -2525,6 +2525,8 @@ confined to the active pane bounds.
   modal is open the paste is swallowed so it can never leak into the
   terminal in the pane behind the overlay; otherwise it pastes into
   the active PTY.
+- **`Ctrl+V` with an image on the clipboard**: handed to the agent in
+  the pane instead of pasted — see "Pasting images" below.
 - **`Ctrl+Shift+V`** (your terminal's paste): the way to paste when
   thurbox runs over SSH — see "Pasting over SSH" below.
 - **`Cmd+C` / `Cmd+V`** (macOS): the same two actions, declared beside the
@@ -2620,6 +2622,23 @@ for one can stall for seconds. When no local clipboard is reachable,
 `Ctrl+V` shows a hint pointing at your terminal's own paste
 (usually **`Ctrl+Shift+V`**), which delivers the text as an ordinary
 bracketed paste that thurbox routes exactly like `Ctrl+V`.
+
+### Pasting images
+
+thurbox pastes text. An image on the clipboard is handed to the **agent**
+instead: `Ctrl+V` is sent to the pane as-is, and a CLI that knows how to read
+the clipboard itself picks the image up from there (Claude Code shells out to
+`xclip`/`wl-paste`, or to PowerShell under WSL). Swallowing the press instead
+is what used to make pasting a screenshot do nothing at all.
+
+**Inside WSL this needs asking Windows.** WSLg bridges the clipboard's *text*
+only: copy a screenshot in Windows and the Linux side is not updated — it keeps
+handing out whatever text was copied before, so a paste inserts something stale
+rather than the image. thurbox therefore asks `powershell.exe` whether the
+Windows clipboard holds an image before deciding what `Ctrl+V` means. The call
+costs about 0.4 s, so it runs **on a worker**: the interface keeps drawing and
+the paste lands when the answer does. Nothing is asked off WSL, where the local
+clipboard is the one being copied into.
 
 ### Pasting on Windows
 
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -2637,8 +2637,12 @@ handing out whatever text was copied before, so a paste inserts something stale
 rather than the image. thurbox therefore asks `powershell.exe` whether the
 Windows clipboard holds an image before deciding what `Ctrl+V` means. The call
 costs about 0.4 s, so it runs **on a worker**: the interface keeps drawing and
-the paste lands when the answer does. Nothing is asked off WSL, where the local
-clipboard is the one being copied into.
+the paste lands when the answer does — in the pane the press was aimed at, not
+whichever one is focused by then. One question runs at a time (a held `Ctrl+V`
+would otherwise start a PowerShell per repeat), presses made while it is out are
+answered by a question of their own, and a question that has not come back in
+five seconds is abandoned and read as "no image". Nothing is asked off WSL,
+where the local clipboard is the one being copied into.
 
 ### Pasting on Windows
 
--- a/docs/FEATURES.md
+++ b/docs/FEATURES.md
@@ -2641,8 +2641,21 @@ the paste lands when the answer does — in the pane the press was aimed at, not
 whichever one is focused by then. One question runs at a time (a held `Ctrl+V`
 would otherwise start a PowerShell per repeat), presses made while it is out are
 answered by a question of their own, and a question that has not come back in
-five seconds is abandoned and read as "no image". Nothing is asked off WSL,
-where the local clipboard is the one being copied into.
+five seconds is abandoned — an unanswerable question is handed to the agent, not
+read as "no image", because the text it would paste instead is the stale one.
+Nothing is asked off WSL, where the local clipboard is the one being copied
+into.
+
+**On macOS the byte is synthesised.** `Cmd+V` is the paste binding there, and a
+`Cmd` chord has no pty encoding at all — handing it on by declining it would
+drop it — so thurbox sends the literal `Ctrl+V` byte the agent watches for.
+Written from the encoding rules rather than from a Mac: the decision is covered
+by a test, the round trip on real hardware is not.
+
+**What the agent reads, thurbox has not seen.** This is the one paste path where
+what you copied reaches the agent as content it fetches itself, rather than as
+text thurbox brackets and sends — and an image carries instructions as readily
+as text does. The press is yours, but the content arrives unread.
 
 ### Pasting on Windows
 
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -52,8 +52,9 @@
 //! arrives as an ordinary bracketed paste.
 
 use std::io::Write;
-use std::process::{Command, Stdio};
+use std::process::{Child, Command, Stdio};
 use std::sync::mpsc::{channel, Receiver, Sender};
+use std::time::{Duration, Instant};
 
 use base64::Engine as _;
 
@@ -270,6 +271,11 @@ pub const PASTE_UNAVAILABLE_HINT: &str =
 #[derive(Default)]
 pub struct ImageProbe {
     channel: Option<(Sender<bool>, Receiver<bool>)>,
+    /// Whether a question is out. One at a time: the clipboard does not change
+    /// between two presses a fifth of a second apart, so a second `powershell.exe`
+    /// would buy nothing and cost another cold start — and key auto-repeat can
+    /// hold `Ctrl+V` down, which without this is a process per repeat.
+    in_flight: bool,
 }
 
 /// How PowerShell is found. `powershell.exe` is on `PATH` inside a distro
@@ -296,10 +302,16 @@ impl ImageProbe {
 
     /// Ask Windows, on a thread. The answer arrives at a later [`Self::poll`].
     ///
-    /// Repeated presses each ask: two pastes are two pastes, and the answers
-    /// come back in the order the threads finish, which for a question this
-    /// small is the order they were asked.
+    /// A press made while an earlier question is still out does **not** ask
+    /// again: one answer serves every press waiting on it, because what is being
+    /// asked about cannot have changed in the time between them. Which presses
+    /// those were is the caller's to remember — this end of the wire only knows
+    /// what the clipboard holds.
     pub fn ask(&mut self) {
+        if self.in_flight {
+            return;
+        }
+        self.in_flight = true;
         let tx = self.channel.get_or_insert_with(channel).0.clone();
         std::thread::spawn(move || {
             let _ = tx.send(windows_clipboard_has_image());
@@ -307,9 +319,15 @@ impl ImageProbe {
     }
 
     /// The answer to one earlier [`Self::ask`], if one has come back.
+    ///
+    /// Taking the answer is what frees the next question: a probe that is never
+    /// polled is never re-asked, which is the behaviour that keeps one wedged
+    /// `powershell.exe` from becoming one per press.
     pub fn poll(&mut self) -> Option<bool> {
         let (_, rx) = self.channel.as_ref()?;
-        rx.try_recv().ok()
+        let answer = rx.try_recv().ok()?;
+        self.in_flight = false;
+        Some(answer)
     }
 }
 
@@ -317,20 +335,65 @@ impl ImageProbe {
 /// the question could not be put: a machine that cannot answer is one whose
 /// clipboard thurbox pastes as text, which is what it did before this existed.
 fn windows_clipboard_has_image() -> bool {
-    let ask = |exe: &str| {
-        Command::new(exe)
+    let ask = |exe: &str| -> std::io::Result<bool> {
+        let mut child = Command::new(exe)
             .args(["-NoProfile", "-NonInteractive", "-Sta", "-Command"])
             .arg(CONTAINS_IMAGE)
             .stdin(Stdio::null())
             .stdout(Stdio::null())
             .stderr(Stdio::null())
-            .status()
+            .spawn()?;
+        // Past the spawn, every way this can go wrong is still just "no image":
+        // the fallback below is for a `powershell.exe` that could not be *found*,
+        // and running a second one because the first became unreadable would be
+        // two wedged processes instead of one.
+        Ok(wait_bounded(&mut child, PROBE_TIMEOUT))
     };
     match ask(POWERSHELL) {
-        Ok(status) => status.success(),
+        Ok(answer) => answer,
         Err(e) => {
             tracing::debug!("{POWERSHELL} is not on PATH ({e}); trying the interop path");
-            ask(POWERSHELL_FALLBACK).is_ok_and(|status| status.success())
+            ask(POWERSHELL_FALLBACK).unwrap_or(false)
+        }
+    }
+}
+
+/// How long the probe waits for Windows before giving up on it.
+///
+/// The round trip is ~0.42 s measured, so this is not a budget anyone reaches
+/// by being slow — it is there because WSL interop can wedge outright (this
+/// machine has logged `UtilAcceptVsock:273: accept4 failed 110`), and a wedged
+/// question with no deadline is a `powershell.exe` and a thread that never end,
+/// one per press, for as long as the interface runs.
+const PROBE_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// How often the wait looks. Next to a 0.42 s round trip this is noise, and it
+/// costs nothing while the probe is out: the thread is asleep.
+const PROBE_POLL: Duration = Duration::from_millis(25);
+
+/// Whether `child` exited successfully within `timeout`, killing it if not.
+///
+/// `false` for a child that overran, was killed, or became unreadable — the
+/// answer that keeps thurbox doing what it did before the probe existed.
+fn wait_bounded(child: &mut Child, timeout: Duration) -> bool {
+    let deadline = Instant::now() + timeout;
+    loop {
+        match child.try_wait() {
+            Ok(Some(status)) => return status.success(),
+            Ok(None) if Instant::now() >= deadline => {
+                tracing::warn!(
+                    "the Windows clipboard probe did not answer within {timeout:?}; \
+                     treating the clipboard as text"
+                );
+                let _ = child.kill();
+                let _ = child.wait();
+                return false;
+            }
+            Ok(None) => std::thread::sleep(PROBE_POLL),
+            Err(e) => {
+                tracing::debug!("could not wait for the Windows clipboard probe: {e}");
+                return false;
+            }
         }
     }
 }
@@ -394,6 +457,86 @@ mod tests {
         assert!(
             windows_clipboard_has_image(),
             "PowerShell says the clipboard holds an image and the probe did not"
+        );
+    }
+
+    /// A probe that never answers is killed, not waited on.
+    ///
+    /// WSL interop wedges (`UtilAcceptVsock:273: accept4 failed 110` on this
+    /// machine): the child is started and simply never returns. Without a
+    /// deadline the worker thread and its `powershell.exe` live for the rest of
+    /// the session — and because the probe is only re-asked once its answer is
+    /// taken, the paste chord would stop being answered at all from then on.
+    ///
+    /// Driven with a real child that really hangs, because the claim is about
+    /// the waiting: a helper asked to time out against nothing proves nothing.
+    #[cfg(unix)]
+    #[test]
+    fn a_probe_that_never_answers_is_killed_rather_than_waited_on() {
+        let mut child = Command::new("sh")
+            .args(["-c", "sleep 30"])
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .spawn()
+            .expect("spawn a hanging child");
+
+        let started = Instant::now();
+        let answered = wait_bounded(&mut child, Duration::from_millis(200));
+        let waited = started.elapsed();
+
+        assert!(!answered, "a child that never exited was read as an answer");
+        assert!(
+            waited < Duration::from_secs(5),
+            "the wait ran past its deadline: {waited:?}"
+        );
+        // Killed, not merely abandoned: an abandoned one holds the interop pipe.
+        assert!(
+            matches!(child.try_wait(), Ok(Some(_))),
+            "the child outlived the wait that gave up on it"
+        );
+    }
+
+    /// A press made while a question is out does not ask a second one.
+    ///
+    /// Key auto-repeat holds `Ctrl+V` down at tens of presses a second, and each
+    /// one used to spawn its own `powershell.exe` — ~0.42 s of cold start each,
+    /// all asking the same question about a clipboard that cannot have changed
+    /// in between. One question, one answer, however many presses are waiting
+    /// on it.
+    ///
+    /// Asserted by counting answers rather than processes: a second question
+    /// would deliver a second `true`/`false` down the same channel, so a probe
+    /// that answers exactly once is a probe that asked exactly once. Runs
+    /// anywhere — off WSL the command cannot start and the failure is the
+    /// answer, which is the same one answer.
+    #[test]
+    fn one_question_is_asked_at_a_time() {
+        let mut probe = ImageProbe::default();
+        probe.ask();
+        probe.ask();
+
+        let deadline = Instant::now() + Duration::from_secs(30);
+        let mut first = None;
+        while first.is_none() && Instant::now() < deadline {
+            first = probe.poll();
+            if first.is_none() {
+                std::thread::sleep(Duration::from_millis(20));
+            }
+        }
+        assert!(
+            first.is_some(),
+            "the probe never answered at all, so this asserts nothing about a \
+             second one"
+        );
+
+        // A second question was asked at the same moment as the first, so its
+        // answer would already be queued behind the one just taken.
+        std::thread::sleep(Duration::from_millis(500));
+        assert_eq!(
+            probe.poll(),
+            None,
+            "a press made while a question was out asked Windows again"
         );
     }
 
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -455,7 +455,7 @@ fn wait_bounded(child: &mut Child, timeout: Duration) -> bool {
             Ok(None) if Instant::now() >= deadline => {
                 tracing::warn!(
                     "the Windows clipboard probe did not answer within {timeout:?}; \
-                     treating the clipboard as text"
+                     what the clipboard holds stays unknown and the press goes to the agent"
                 );
                 let _ = child.kill();
                 let _ = child.wait();
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -541,10 +541,11 @@ mod tests {
     fn only_a_wsl_distro_asks_windows() {
         use crate::session::host_def::with_wsl_distro;
 
-        assert!(
+        assert_eq!(
             with_wsl_distro(Some("Ubuntu"), ImageProbe::applies),
+            cfg!(unix),
             "inside a distro the X clipboard is not the one being copied into, \
-             so Windows has to be asked"
+             so Windows has to be asked — and a native Windows build asks nobody"
         );
         assert!(
             !with_wsl_distro(None, ImageProbe::applies),
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -52,6 +52,7 @@
 //! arrives as an ordinary bracketed paste.
 
 use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
 use std::process::{Child, Command, Stdio};
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::time::{Duration, Instant};
@@ -283,6 +284,23 @@ pub struct ImageProbe {
 const POWERSHELL: &str = "powershell.exe";
 const POWERSHELL_FALLBACK: &str = "/mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe";
 
+/// Where PowerShell is looked for, in the order it is tried.
+///
+/// The `PATH` search goes through [`crate::paths::resolve_on_path`] rather than
+/// through `Command::new`'s own: the OS search treats an **empty** entry in
+/// `PATH` (a stray leading, trailing or doubled `:`) as the current directory,
+/// so a file named `powershell.exe` sitting in whatever repository thurbox was
+/// launched from would answer this question on every `Ctrl+V`. `resolve_on_path`
+/// keeps absolute entries only, which is the rule #1100 landed for agent
+/// spawns; there is no reason for this path to hold a weaker one.
+fn powershell_candidates() -> Vec<PathBuf> {
+    let mut found: Vec<PathBuf> = crate::paths::resolve_on_path(POWERSHELL)
+        .into_iter()
+        .collect();
+    found.push(PathBuf::from(POWERSHELL_FALLBACK));
+    found
+}
+
 /// What Windows said about its clipboard.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Verdict {
@@ -372,7 +390,7 @@ impl ImageProbe {
 /// apart. Every way the question could not be put — spawn, deadline, an answer
 /// that is neither word — is [`Verdict::Unknown`], never `NotImage`.
 fn windows_clipboard_has_image() -> Verdict {
-    let ask = |exe: &str| -> std::io::Result<Verdict> {
+    let ask = |exe: &Path| -> std::io::Result<Verdict> {
         let mut child = Command::new(exe)
             .args(["-NoProfile", "-NonInteractive", "-Sta", "-Command"])
             .arg(CLIPBOARD_KIND)
@@ -402,13 +420,13 @@ fn windows_clipboard_has_image() -> Verdict {
             }
         })
     };
-    match ask(POWERSHELL) {
-        Ok(verdict) => verdict,
-        Err(e) => {
-            tracing::debug!("{POWERSHELL} is not on PATH ({e}); trying the interop path");
-            ask(POWERSHELL_FALLBACK).unwrap_or(Verdict::Unknown)
+    for exe in powershell_candidates() {
+        match ask(&exe) {
+            Ok(verdict) => return verdict,
+            Err(e) => tracing::debug!("{} could not be run ({e})", exe.display()),
         }
     }
+    Verdict::Unknown
 }
 
 /// How long the probe waits for Windows before giving up on it.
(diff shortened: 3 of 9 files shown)


spscream:
All three review findings are addressed — `95573b8` for the first two, `e4706d3` for the third, `f2f6ea3` for the docs that described the old behaviour.

**The answer went to whatever was focused when it landed.** That was deliberate ("paste where the person is looking") and wrong: the question takes ~0.42 s, which is long enough to switch panes, and a paste arriving in a pane nobody aimed it at corrupts what is being typed there — the same failure as pasting stale text, from the other side. The target session is now taken at the press (`paste_targets`) and passed to the delivery, which no longer reads the focus at all; `on_paste`'s tail was split out as `paste_text_into(session, text)` so a delayed paste can name its own destination.

**The probe had no bound.** Key auto-repeat holds `Ctrl+V` down at tens of presses a second against a question that takes a fifth of one, so that was a `powershell.exe` per repeat. One question is out at a time now, and the queue of presses waiting on one is capped (8 — well above two deliberate pastes, well below what a repeat makes).

**The probe had no deadline.** WSL interop can wedge outright (this machine has logged `UtilAcceptVsock:273: accept4 failed 110`), and a wedged probe with no timeout is a thread and a process that never end — and, since the next question waits on this one's answer, a paste chord that is never answered again. `wait_bounded` kills the child after five seconds and reads it as "no image", which is what thurbox did before the probe existed.

**And the follow-up finding: one answer must not classify presses that postdate it.** With the in-flight bound as first written, a press made while a question was out joined that question's answer — but the clipboard can change while it is out, so an image copied in between would be classified by what preceded it and pasted as stale text. `ask()` now reports whether it actually started a question, and the coordinator records how many presses that question was asked *for*; the rest stay queued and get a question of their own when it returns. Still one process at a time: a held key costs one question per round trip rather than one per repeat.

Tests, each checked with the fix stubbed out:

- `a_probe_that_never_answers_is_killed_rather_than_waited_on` — a real child that really hangs. Without the kill: "the child outlived the wait that gave up on it"; without the deadline: "a child that never exited was read as an answer".
- `one_question_is_asked_at_a_time` — counts answers rather than processes (a second question would deliver a second answer down the same channel), and asserts that taking an answer frees the next question. Without the bound: "a press made while a question was out asked Windows again".

The delivery routing itself has no test, and I could not write an honest one: `App` lives in the binary crate and is built as a single literal in `coordinator/boot.rs`, so nothing in the suite constructs one. What I did instead was make the destination an argument rather than a read of the focus, so the routing is visible in the signature. If there is a harness for driving the coordinator that I missed, I would rather cover it than argue it.

spscream:
Correction to my last comment, and four more fixes from an independent review of the branch (`657aff0`).

**I was wrong that the routing could not be covered.** `tests/tui_e2e.rs` drives the real binary in a pty and lets a test set its environment, which is all this needs: `WSL_DISTRO_NAME=Ubuntu` plus a directory in front of `PATH` holding a `powershell.exe` that counts its calls and answers on stdout. No Windows involved, and no dependence on what the clipboard happens to hold. `a_paste_under_wsl_asks_windows_once_per_press_and_never_from_a_float` now covers the whole route — the press claimed, the question asked on a worker, the answer polled back on the loop and spent, and only then the next question. Deleting `poll_image_probe` from the loop fails it on the second press; that deletion left all 15 unit tests green.

**A float's paste was swallowed.** The clipboard stage runs before `dispatch_grabbed`, so under WSL the press was claimed and a question asked while the new-session wizard held the keyboard — and the answer, arriving 0.42 s later, found the float still up and dropped it. Pasting a repository path into the wizard did nothing at all, with no toast and no error; off WSL the same press typed into the field. Nothing is asked while an overlay owns typed input now: a name field cannot take a picture, so the question has no answer worth 0.42 s. The guard on *delivery* goes the other way and is gone — a press that already named its destination is not a leak past an overlay that went up since, it is the thing that was asked for.

**An ordinary rich copy was read as an image.** Excel, Word, Outlook and browsers put a bitmap on the clipboard alongside the text on a normal copy, so `ContainsImage` alone is true for copying a spreadsheet row: thurbox handed the press to the agent, which fetched a picture of the row the person meant to paste as text. Before this PR WSLg bridged the text and `Ctrl+V` pasted it correctly, so that was a regression on a common path. `ContainsText` is the tie-break now — only a clipboard carrying a picture and no text is an image paste.

**A wedged interop was read as "no image".** PowerShell exits non-zero for a clipboard without a picture, a missing assembly and a wedged `UtilAcceptVsock` alike, and only the first means "paste the text". The reviewer measured the wedge on this machine at roughly one cold call in ten (`UtilAcceptVsock:273: accept4 failed 110`, rc=1, empty stdout), each of which pasted the stale X-clipboard text this path exists to stop. The answer is a word on stdout now, and a question that could not be put becomes `Verdict::Unknown`, which is given to the agent rather than folded into "no image". The oracle test was skipping on exactly that failure and passing — it now separates "PowerShell could not be asked" from "the clipboard holds no image", and asserts on both halves of the tie-break.

**A paste could land where a keystroke would not.** The chord went to `focused_session` while `dispatch_session_input` routes by `focused_surface`, and a program pane's keys go to the program: with an editor pane focused, the same press reached the program off WSL and the shell under it on WSL. Both paste paths route by surface through one helper now, so a pane that takes typing takes pastes.

Suite: 2610 run, the same four pre-existing environment-dependent failures on this machine.

spscream:
**First, a correction: `657aff0` was not on this branch when I wrote about it.**
The push never landed, so the four fixes my last comment described — the float
that lost its paste, the `Verdict::Unknown` split, routing by surface, and the
e2e test that drives the real binary — were sitting on a local branch while the
PR still showed `f2f6ea3`. They are pushed now, together with `a21ca00`.

On the comment finding: the half that matters is accepted, and it was worse than
verbosity — two of those docs were **stale**, which this repository rates as
worse than no comment at all. `windows_clipboard_has_image` was documented as
returning `false` for "no image" when it returns one of three verdicts, and
telling them apart is the whole point of the commit that introduced it;
`wait_bounded` claimed its `false` kept thurbox "doing what it did before the
probe existed", which is the opposite of what a killed probe now does — the
press goes to the agent precisely so the stale X-clipboard text is not pasted.

Trimmed with them: the one-question-at-a-time rule, which was argued three times
over (the field, `ask`, `poll`) and is now stated once where the flag lives; and
the routing rule, which was written out twice because `dispatch_session_input`
inlined the same match `send_to_surface` performs — the keystroke path calls the
helper now, so the rule has one home and a paste cannot drift from a keystroke.

What stays is what a reader cannot infer from the code, and what the finding
itself said to keep: the measurements (~0.42 s per round trip, WSLg bridging
text only), the `ContainsText` tie-break that keeps an ordinary Excel copy from
being read as an image, and why an unanswerable question must not become "no
image".

src/coordinator/input.rs
```diff
@@ -591,43 +584,229 @@ impl App {
...
+    /// terminals disable clipboard *reads* by default and probing for one can
+    /// stall for seconds.
+    fn paste_text_or_decline(&mut self) -> bool {
+        if self.clipboard.is_none() {
             self.toast(thurbox::clipboard::PASTE_UNAVAILABLE_HINT);
-            return;
+            return true;
+        }
+        let Some(text) = thurbox::clipboard::paste(self.clipboard.as_mut()) else {
+            return false;
```
LeTuR: On macOS this decline drops the press rather than handing it on, so the bug in the title is unfixed there.

`cmd+v` is a declared paste binding (`kernel::clipboard::CLIPBOARD`, added on macOS builds), and `agent::input::key_to_bytes` returns `None` for anything carrying `SUPER` — "Cmd/Super-modified keys have no legacy PTY encoding". So a declined `Cmd+V` walks the rest of `on_key`, reaches `dispatch_session_input`, gets `None` back and is dropped. `Ctrl+V` on the same Mac works; the chord a Mac user actually presses still does nothing on a copied image. Before this change they at least got the (misleading) clipboard hint — now it is silence.

`deliver_probed_paste` already solves this the other way round: it sends the literal `Ctrl+V` byte instead of relying on a fall-through. Would you do the same here when the press carries `SUPER`? If you would rather not guess at macOS without a machine to check on, that is a fair answer — but then say so in the new docs/FEATURES.md section, so it does not claim a platform it does not cover.

src/clipboard.rs
```diff
@@ -238,10 +241,401 @@ pub fn paste(native: Option<&mut arboard::Clipboard>) -> Option<String> {
...
+    /// Whether a question is out. At most one at a time, because key
+    /// auto-repeat holds `Ctrl+V` down far faster than the ~0.42 s answer and
+    /// a process per repeat is a machine brought to its knees by a held key.
+    in_flight: bool,
+}
+
+/// How PowerShell is found. `powershell.exe` is on `PATH` inside a distro
+/// through WSL interop; the absolute path is the fallback for a `PATH` that
+/// interop did not reach, and is the one Claude Code itself falls back to.
+const POWERSHELL: &str = "powershell.exe";
```
LeTuR, line 285: Security: `Command::new("powershell.exe")` resolves through the OS `PATH` search, which reads an empty `PATH` entry (`:/usr/bin`, or a stray trailing colon) as the current directory — so a file named `powershell.exe` sitting in the repository thurbox was launched from can be what answers, on every `Ctrl+V`. That is the same hole #1100 (dd7faf3) closed for agent spawns, and the rule it landed was "absolute `PATH` entries only".

It is bounded: the child gets a null stdin, its stderr is discarded, and only two exact words are believed. It is still an attacker-placed binary being executed.

`paths::resolve_on_path` already implements the absolute-only rule — though `clipboard`'s `ModuleRules` entry in tests/architecture_rules.rs allows `session` only, so it would need `paths` added beside it. The cheaper alternative is to try `POWERSHELL_FALLBACK`'s absolute path first and only fall back to the bare name. Either way the e2e test needs another way to install its stub, since it relies on this bare-name lookup.

src/clipboard.rs
```diff
@@ -238,10 +241,401 @@ pub fn paste(native: Option<&mut arboard::Clipboard>) -> Option<String> {
...
+        Ok(match answer.trim() {
+            IMAGE_ANSWER => Verdict::Image,
+            OTHER_ANSWER => Verdict::NotImage,
+            other => {
+                tracing::debug!("the Windows clipboard probe answered {other:?}");
+                Verdict::Unknown
+            }
+        })
+    };
+    match ask(POWERSHELL) {
```
LeTuR: Where PowerShell cannot be reached at all, this returns `Unknown` for every press for the rest of the session, and `deliver_probed_paste` turns each one into a forwarded `Ctrl+V` — so ordinary *text* paste stops working under WSL on any box where neither candidate exists: interop off in `wsl.conf`, `automount.enabled = false`, or a locked-down image. thurbox never pastes, and an agent that is not Claude Code does nothing with the chord.

"Interop wedged this once" and "there is no PowerShell on this machine" are different facts and only the first is worth retrying. Could two `NotFound` spawn errors latch a "cannot ask here" flag on `ImageProbe`, so `applies()` goes false and `paste_text_or_decline` takes over for the rest of the session? A timeout or an unrecognised answer would stay `Unknown` exactly as it is now, which is the behaviour the doc comment on `Verdict::Unknown` argues for.

src/coordinator/input.rs
```diff
@@ -591,43 +584,229 @@ impl App {
...
+    /// exists to stop. The hand-off is a literal `Ctrl+V` byte rather than a
+    /// fall-through to the chord, since by now the key press is long gone; that
+    /// byte is what Claude Code watches for before reading the clipboard itself
+    /// (`xclip`/`wl-paste`, or PowerShell under WSL).
+    ///
+    /// Delivered even if a modal or a float has gone up since: this press
+    /// already named its destination, so putting it there is not a leak past
+    /// the overlay but the thing that was asked for. What an overlay prevents
+    /// is the *question* — see [`Self::paste_into_focused`].
+    fn deliver_probed_paste(&mut self, verdict: thurbox::clipboard::Verdict, surface: &str) {
```
LeTuR, line 821: This delivers into the pane behind an overlay that went up while the question was out, and docs/FEATURES.md states the opposite rule for the press itself: "while a modal is open the paste is swallowed so it can never leak into the terminal in the pane behind the overlay". Press `Ctrl+V`, then `Ctrl+N` within the ~0.42 s, and a bracketed paste — or a bare `Ctrl+V` — lands on the agent behind the wizard.

Your doc comment argues the press named its destination before the overlay existed, and I can see it; I am not sure which way I would rule either. What I do not want is the two paragraphs contradicting each other for the next reader. Which is it meant to be, and can the losing side get a line saying it is deliberate?

tests/tui_e2e.rs
```diff
@@ -1266,3 +1272,123 @@ fn the_scrollbar_can_be_pressed_and_dragged() {
...
+///   never to be asked again — but the interface then swallows every paste for
+///   the rest of the session. Here the second press has to reach Windows, which
+///   it can only do after the first answer was taken.
+/// - **a float's paste is swallowed.** The clipboard stage runs *before*
+///   `dispatch_grabbed`, so asking Windows while the new-session wizard is up
+///   claims a press that the float should have had — and the answer, arriving
+///   0.42 s later, finds the float still there and drops it. Pasting a path
+///   into the wizard did nothing at all under WSL.
+#[test]
+fn a_paste_under_wsl_asks_windows_once_per_press_and_never_from_a_float() {
```
LeTuR, line 1341: Everything with a test here is the WSL half. The change that reaches every user is the other one — `paste_into_focused` returning `false` when a reachable clipboard holds no text, so the press falls through to the pane — and nothing pins it. Put `Some(true)` back in `run_clipboard_action` and this suite stays green on Linux, macOS and Windows, which is where the reported bug was.

A true e2e is awkward (you would need arboard present and empty), so I would take a unit test over the decision instead: a function of "is there a clipboard" and "did it yield text" that answers claim-or-decline, with `paste_text_or_decline` calling it. One test, not five.

docs/FEATURES.md
```diff
@@ -2621,6 +2623,27 @@ for one can stall for seconds. When no local clipboard is reachable,
 (usually **`Ctrl+Shift+V`**), which delivers the text as an ordinary
 bracketed paste that thurbox routes exactly like `Ctrl+V`.
 
+### Pasting images
```
LeTuR, line 2629: nit, security: this is the first path where something the user copied from elsewhere — a web page, a chat, a README — reaches the agent as content the agent fetches and reads itself, instead of as text thurbox brackets and sends. An image carries instructions as readily as text does, and the agent cannot tell a pasted screenshot from the user's own prompt.

That is a property of handing the chord over, not a defect in this code, and the user did press paste. I am not asking for a behaviour change. One sentence in this section saying the content goes to the agent unread by thurbox would put the trust story next to the feature.

LeTuR (changes requested):
Read the whole diff and the code around it. Since CI does not run on forks I ran the gate locally against a21ca00: `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, `cargo deny check advisories` and `cargo nextest run --all` are all clean — 2611 tests pass, including your three unit tests and the new e2e (2.7 s here on Linux, with the stub standing in for Windows).

The shape of the fix is the right one. Declining the chord rather than swallowing it is the correct answer and it is the one `kernel.copy` already gives; asking Windows off the loop with the target pinned at the press, one question at a time and a deadline on the child, is the part I would otherwise have had to ask for.

Two things I need changed before this goes in:

- **`Cmd+V` on macOS drops the press** instead of handing it on, so the bug in the title is unfixed on the chord a Mac user presses. Comment on the decline in `paste_text_or_decline`.
- **The probe resolves `powershell.exe` through the OS `PATH` search**, which can execute a binary out of the working directory. #1100 closed exactly this for agent spawns two weeks ago.

Three more inline that are questions rather than demands: `Verdict::Unknown` permanently disabling text paste on a WSL box without reachable interop; a probe answer landing behind an overlay, against what FEATURES.md says about modals; and no test for the non-WSL half of the change.

Who does what: all of it is yours — I cannot push to your fork. Nothing here needs a follow-up issue; it all fits in this branch. Ping me when it is updated and I will re-read.

— LeTuR's agent, reviewing as maintainer

spscream:
Both blockers are fixed; the three questions are answered below, two of them
with "deferred, and here is what I would pick".

**CI was one root cause.** `cog.toml` allows `api, cli, ui, git, core, docs,
deps, config, mcp` — this branch used `clipboard`, so all six commits and the
title failed the conventional-commit gate, and Nextest / Windows were
*cancelled* rather than failing. Everything is `fix(core)`/`docs(core)` now, the
scope `main` uses for the neighbouring work. Force-pushed, so the hashes moved.

**`Cmd+V` on macOS — fixed.** You are right about the mechanism: `key_to_bytes`
refuses every `SUPER` chord, so declining a press there hands it to nothing. A
declined paste the pty cannot carry now sends the literal `Ctrl+V` byte itself,
the way `deliver_probed_paste` already does, and only where the fall-through
would have delivered anyway — no overlay owning typed input, and a focused pane
that asked for raw session input. Written from the encoding rules, not from a
Mac: `docs/FEATURES.md` says exactly that, so the section does not claim a
round trip nobody ran.

**PowerShell on `PATH` — fixed through `paths::resolve_on_path`**, the rule
#1100 landed, with `clipboard` gaining `paths` in the module allowlist. The
hard-coded interop path is now the last *candidate* rather than a retry after a
failed spawn, so a post-spawn failure is still `Unknown` with no second wedged
process. The e2e stub needed nothing: its directory is absolute, which is
precisely what the rule keeps. Tested where it bites — a `powershell.exe`
planted under a relative `PATH` entry that really does resolve from the working
directory must not be a candidate; the OS search picks it up, and that is what
fails when the resolver is swapped back.

**The untested half — done, as the one test you asked for.** `paste_route` is
the decision over your two facts, and both callers read it: the plain press and
the one that waited on Windows, whose verdict says only whose press it is. Put
`Some(true)` back in `run_clipboard_action` and that test fails.

**The overlay question — I would rule for delivery, but it is yours to settle.**
The press named its destination before the overlay existed, and cancelling it on
focus change is the version I would want if we go the other way. What I am not
doing is leaving the two paragraphs contradicting each other: tell me which way
you want it and the losing side gets the line saying it is deliberate. Greptile
filed the same thing as a P1 on the same lines, for what that is worth.

**The `Unknown` latch — agreed in principle, deferred with the overlay call.**
"Interop wedged once" and "there is no PowerShell here" are different facts and
your split is the right one; two `NotFound` spawn errors latching `applies()`
to false is what I would write. It is a behaviour change to the path the other
question is about, so I would rather land it in the same push as that decision
than guess at both.

**The trust note — added**, in your words: what the agent fetches arrives unread
by thurbox, and an image carries instructions as readily as text.

Local gate here: `fmt`, `clippy --all-targets`, 2615 tests green bar four that
fail identically on `main` in this environment. Ready for another read.

spscream:
Rebased on `main` (through #1110) and force-pushed — no conflicts, and the gate
re-run on the rebase is clean: `fmt`, `clippy --all-targets`, 2615 tests bar the
four that fail identically on `main` in this environment. The review bot found
nothing new on the rebased head.

Note for whenever you look again: each force-push leaves the workflow runs at
`action_required`, so Actions has not run since your approval this morning.

spscream:
Confirmation from real hardware, since everything on this branch so far was a
stub: this build is running on a WSL box now, and pasting an image copied in
Windows into the agent works — real `powershell.exe`, real interop, the agent
fetching the picture itself off the forwarded `Ctrl+V`.

That is the path the e2e test can only approximate. It does not cover the macOS
chord, which is still written from the encoding rules and says so in
`docs/FEATURES.md`.

src/clipboard.rs
```diff
@@ -238,10 +242,475 @@ pub fn paste(native: Option<&mut arboard::Clipboard>) -> Option<String> {
...
+fn wait_bounded(child: &mut Child, timeout: Duration) -> bool {
+    let deadline = Instant::now() + timeout;
+    loop {
+        match child.try_wait() {
+            Ok(Some(status)) => return status.success(),
+            Ok(None) if Instant::now() >= deadline => {
+                tracing::warn!(
+                    "the Windows clipboard probe did not answer within {timeout:?}; \
+                     treating the clipboard as text"
+                );
```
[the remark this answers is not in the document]
spscream: Right, and it is the same stale text as the two comments fixed in faf8636 — this one is a log line, so it survived that sweep. `wait_bounded` returning `false` becomes `Verdict::Unknown`, and `Unknown` hands the press to the agent; nothing is pasted as text. Reworded in 0c7229d.

src/coordinator/input.rs
```diff
@@ -591,43 +584,229 @@ impl App {
...
+            }
+            if let Some(text) = thurbox::clipboard::paste(self.clipboard.as_mut()) {
+                self.paste_text_into(surface, &text);
+                return;
+            }
+        }
+        const CTRL_V: u8 = 0x16;
+        if !self.send_to_surface(surface, vec![CTRL_V]) {
+            self.toast("no live terminal to paste into");
+        }
```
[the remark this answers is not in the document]
spscream: Same finding as the maintainer's thread on the lines just above, and it is the open question on this PR rather than an oversight: the press named its destination before the overlay existed, so delivering it is arguably the thing that was asked for, not a leak past the overlay.

The guard sits on the *question* instead — no probe is started while an overlay owns typed input. What is not settled is delivery, and I would take cancel-on-focus-change if the call goes the other way. Left open for the maintainer to rule.

src/coordinator/input.rs
```diff
@@ -591,43 +584,229 @@ impl App {
...
+        use thurbox::clipboard::Verdict;
+        if verdict == Verdict::NotImage {
+            if self.clipboard.is_none() {
+                self.toast(thurbox::clipboard::PASTE_UNAVAILABLE_HINT);
+                return;
+            }
+            if let Some(text) = thurbox::clipboard::paste(self.clipboard.as_mut()) {
+                self.paste_text_into(surface, &text);
+                return;
+            }
```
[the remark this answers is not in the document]
spscream: Outdated — the lines moved — but worth answering, because half of it was real. A press made *after* the question was sent used to be answered by it; that is fixed by `probed_presses` (`714a111`), and the target is taken at the press rather than at the answer (`20946ba`).

What remains is narrower than the comment reads: the text is fetched after the verdict, so the only wrong mix is an image copied during the 0.42 s the question was in flight, which is classified as text. Pasting text that is on the clipboard *now* is not stale. Closing that last window means snapshotting both facts atomically, which the interop does not offer.

src/coordinator/input.rs
spscream: Fixed in `54bcd65`. You had the mechanism right: `key_to_bytes` returns `None` for every chord carrying `SUPER`, so declining there handed the press to nothing.

A declined paste the pty cannot carry now sends the literal `Ctrl+V` byte itself, the way `deliver_probed_paste` already does, and only where the fall-through would have delivered anyway — no overlay owning typed input, a focused pane that asked for raw session input. Pinned by `a_cmd_chord_cannot_be_handed_on_by_declining_it`. Written from the encoding rules rather than from a Mac, and `docs/FEATURES.md` says so.

src/clipboard.rs
spscream: Fixed in `7e3e27b`, through `paths::resolve_on_path` — the rule #1100 (dd7faf3) landed for agent spawns — with `clipboard` gaining `paths` in the module allowlist. The hard-coded interop path is now the last *candidate* rather than a retry after a failed spawn, so a post-spawn failure stays `Unknown` with no second wedged process.

Tested where it bites: `the_probe_looks_only_at_absolute_path_entries` plants a `powershell.exe` under a relative `PATH` entry that really does resolve from the working directory, and asserts the candidate list is exactly the interop path. Swap the resolver back for the OS search and that test fails — the first version of it did not, because nothing was planted where the relative entry pointed, and it proved nothing.

src/clipboard.rs
spscream: Agreed in principle and deferred with the overlay call, as set out in https://github.com/Thurbeen/thurbox/pull/1108#issuecomment-5645736942: "interop wedged once" and "there is no PowerShell here" are different facts, and two `NotFound` spawn errors latching `applies()` to false is what I would write.

It changes behaviour on the same path the overlay question is about, so I would rather land both in one push than guess at one of them.

src/coordinator/input.rs
spscream: This is the one I would rather you settled than I did. My reading: a press that already named its destination is not a leak past an overlay that went up since — it is the thing that was asked for; the guard belongs on the question, which is where it is. If you rule the other way, the version I would want is cancel-on-focus-change rather than a silent drop at delivery.

Either way the two paragraphs in `docs/FEATURES.md` stop contradicting each other in the same push: the losing side gets the line saying it is deliberate. Greptile filed the same thing as a P1 on these lines, for what that is worth.

tests/tui_e2e.rs
spscream: Done as the one test you asked for, in `54bcd65`. `paste_route(clipboard_present, yielded_text) -> PasteRoute` is the decision over your two facts, and both callers read it — the plain press and the one that waited on Windows, whose verdict only decides whose press it is:

```rust
fn paste_route(clipboard_present: bool, yielded_text: bool) -> PasteRoute {
    match (clipboard_present, yielded_text) {
        (false, _) => PasteRoute::Hint,
        (true, true) => PasteRoute::Send,
        (true, false) => PasteRoute::GiveToAgent,
    }
}
```

Put `Some(true)` back in `run_clipboard_action` and `a_clipboard_with_nothing_to_paste_hands_the_press_on` fails.

docs/FEATURES.md
spscream: Added in `54bcd65`, in your framing: what the agent fetches off the forwarded `Ctrl+V` arrives unread by thurbox, and an image carries instructions as readily as text — so the section says what this path does and does not check, rather than leaving the reader to assume thurbox saw it.

src/coordinator/input.rs
LeTuR: My call: **deliver, as you have it.** The code stays; the docs get the line.

The two situations are not the same one, which is what I could not settle when I wrote the comment. A press made while an overlay is up never had a destination — FEATURES.md's rule is about that press, and swallowing it is right. A press made before the overlay existed already named a pane, and the answer arriving late does not unname it. So `deliver_probed_paste` is correct and the paragraph that needs the line is the FEATURES.md one: say that the swallowing rule is about the press, and that an answer owed to a press made before the overlay went up is still delivered where it was aimed.

I am not taking the cancel-on-focus-change version, and I would not take it later either: the target is pinned at the press precisely so focus moving does not redirect it, and a paste that vanishes because the person looked elsewhere for a third of a second is the silent loss this whole branch is about.

One thing I would add while you are in there: the hand-off branch sends its byte without a toast, so a `Ctrl+V` forwarded behind an overlay is the one delivery with no trace on screen. `paste_text_into` already toasts. Not a blocker, your judgement on whether it is worth the noise.

src/clipboard.rs
LeTuR: Take it. Two `NotFound` spawn errors latch `applies()` to false for the session, and `paste_text_or_decline` handles the rest — that is the behaviour I want.

One refinement, since you will be writing the discrimination anyway: `NotFound` is the only error that means "there is no PowerShell on this machine". `PermissionDenied`, an exec format error, or anything else is a machine that has one and would not run it this time, and latching on those would turn a transient or policy failure into a permanent one. Keep those at `Verdict::Unknown` with no latch, as now.

The other half of the reason to do it: on a box that latches, `applies()` going false also takes the ~0.42 s off every paste, which is the cost the gate exists to control in the first place.

src/clipboard.rs
```diff
@@ -238,10 +242,475 @@ pub fn paste(native: Option<&mut arboard::Clipboard>) -> Option<String> {
...
+    /// The gate is the whole cost control: every paste on a machine that asks
+    /// pays ~0.42 s for a `powershell.exe` that a plain Linux or macOS box has
+    /// no reason to run. A gate that answered "yes" everywhere would be a
+    /// third-of-a-second added to every `Ctrl+V` on every platform.
+    #[test]
+    fn only_a_wsl_distro_asks_windows() {
+        use crate::session::host_def::with_wsl_distro;
+
+        assert!(
+            with_wsl_distro(Some("Ubuntu"), ImageProbe::applies),
```
LeTuR, line 623: This fails on native Windows, and `Windows (tests)` on this head is red for it alone — 2556 passed, this one failed, three retries:

```
thread 'clipboard::tests::only_a_wsl_distro_asks_windows' panicked at src\clipboard.rs:544:9:
inside a distro the X clipboard is not the one being copied into, so Windows has to be asked
```

`applies()` is `cfg!(unix) && current_wsl_distro().is_some()`, so on a `windows-msvc` build it is false whatever `WSL_DISTRO_NAME` says. The production behaviour is right — native Windows reads its own clipboard through arboard and has no reason to ask — it is the assertion that is unconditional.

The gate itself is worth pinning on both platforms rather than compiling the test away, so I would make the expectation the `cfg`:

```rust
assert_eq!(
    with_wsl_distro(Some("Ubuntu"), ImageProbe::applies),
    cfg!(unix),
    "inside a distro the X clipboard is not the one being copied into, so \
     Windows has to be asked — and a native Windows build asks nobody",
);
```

The second assertion holds as it is. Nothing to do with your last three commits: it has been here since 1cec14c and CI never got to it, because the `clipboard` scope failed the conventional-commit gate and Nextest and both Windows jobs were cancelled rather than run. The scope rename is what let it through to a real runner.

LeTuR (changes requested):
Both my blockers are fixed and I read the fixes. One new thing stands, and it is not from your last three commits: `Windows (tests)` is red.

`clipboard::tests::only_a_wsl_distro_asks_windows` fails on `windows-msvc` because `applies()` is `cfg!(unix) && …` and the assertion is unconditional. Comment on the line with the one-line shape I would take. The production behaviour is correct; it is the test that claims too much. It has been on the branch since 1cec14c and nothing ran it: the `clipboard` scope failed the conventional-commit gate, so Nextest and both Windows jobs were cancelled rather than executed. Renaming the scope is what finally pointed a Windows runner at it — so the title fix earned its keep twice.

Where the rest landed:

- **`Cmd+V` on macOS — fixed** (54bcd65). `hand_unencodable_paste_over` synthesises the byte, gated on `reaches_the_pty` and on the two conditions the fall-through needs anyway, so `Ctrl+V` still travels the old route and only an unencodable chord is synthesised. Pulling the decision into `paste_route` is more than I asked for and the right more: the probed press and the plain one cannot drift apart now. FEATURES.md saying the round trip was never run on a Mac is the honest way to ship an untested platform — I would rather read that than a claim.
- **PowerShell on `PATH` — fixed** (7e3e27b). `resolve_on_path`, absolute entries only, interop path as the last candidate rather than a retry after a failed spawn, allowlist widened to match. `the_probe_looks_only_at_absolute_path_entries` plants the file where a relative entry really does resolve it, which is the only version of that test worth having.
- **No test for the non-WSL half — fixed.** `a_clipboard_with_nothing_to_paste_hands_the_press_on`, over the decision, failing with `Some(true)` put back. One test, as asked.
- **The trust note — added.** "What the agent reads, thurbox has not seen."
- **The overlay contradiction — answered in thread: deliver, as you have it**, and FEATURES.md gets the line. Reasoning there.
- **The `Unknown` latch — answered in thread: take it**, with one refinement on which spawn errors may latch.

Security, re-run over the changed code: nothing new. The probe still reads no image bytes, so there is no temp file, no decode, no unbounded read, no filename off clipboard metadata, and no clipboard content in a log or a toast; the script is still a `const` in a single argv element with no shell, and the child's stdout is bounded by the pipe because it is read after the wait. Two residuals I am explicitly not asking you to fix: an unexpected probe answer is logged whole rather than truncated, and `resolve_on_path` does not care whether the absolute directory it trusts is world-writable — that is #1100's rule reaching its own limit, the same for agent spawns, and not yours here.

What happens next: the Windows test, the FEATURES.md line and the `Unknown` latch, all yours, all in one push. I am requesting changes for the red check rather than the other two — those I would have taken as follow-ups. When it is green I will approve without a fourth read of the same code. I do not merge: this is a fork, and the operator decides when it goes in.

Verified locally at 0c7229d3 on Linux: `fmt`, `clippy --all-targets --all-features -D warnings`, `cargo deny`, `rumdl`, and `cargo nextest run --all` — 2615/2615, including the four that failed in your environment, so those were environmental. Every other CI job on this head is green.

— LeTuR's agent, reviewing as maintainer

spscream:
All three taken, in `653f058`, `8e7ee75` and `7a6aa8e`.

**The Windows red — your version, verbatim.** The production behaviour was right
and the assertion was not, and making the expectation the `cfg` keeps the gate
pinned on both platforms rather than compiling it away on one. You are also
right about why CI never saw it: the `clipboard` scope failed the
conventional-commit gate, so Nextest and both Windows jobs were cancelled rather
than run, and the scope rename is what let this reach a real runner.

**The latch — `NotFound` only, and your refinement is in the code.**
`PermissionDenied`, an exec format error and anything else a spawn can fail with
reset the count rather than feeding it: a machine that has a PowerShell and
would not run it this time must not be latched off permanently. Any answer at
all clears the count too, so what is counted is a standing absence rather than a
total. Two rounds is your number; my only addition is the reason I would not
take one — interop publishes `powershell.exe` on `PATH` a moment after a distro
starts, and a paste in that moment would otherwise silence the probe for the
session. That is caution about an ordering, not a measurement.

The decision is a function of what a round found, so it is tested directly:
absence twice latches and `applies()` goes false inside a distro, refusals never
latch however many there are, and an answer in between makes the count start
again rather than resume. Drop `!powershell_is_absent()` from `applies()` and the
first of those fails.

**The overlay — your call taken as given.** `deliver_probed_paste` stays, and
`docs/FEATURES.md` now says the rule is about the press in both places that
talked about it: the `Ctrl+V` bullet, and a paragraph in "Pasting images" that
says no question is asked while an overlay owns typed input, and that an answer
owed to a press made before the overlay went up is still delivered where it was
aimed. Cancel-on-focus-change is not in the branch and I will not bring it back.

**And the trace you asked about — worth the noise, on that one path.** The
delayed hand-off now toasts: "image left to the agent to fetch", or the
`Unknown` wording when Windows could not be asked. It is the one delivery the
person may not see happen — a byte the agent acts on rather than text appearing
in a prompt, possibly behind an overlay — and the text path already reports
itself the same way. The synchronous `Cmd+V` hand-off stays silent: it cannot
land behind an overlay (`hand_unencodable_paste_over` refuses when one owns
input), and the agent's own reaction is on screen a moment later.

Local gate on `7a6aa8e`: `fmt`, `clippy --all-targets`, 2616 tests green bar the
four that fail identically on `main` in this environment. The review bot found
nothing new on the new head.

src/clipboard.rs
spscream: Taken in `8e7ee75`, with your refinement: only `NotFound` counts, everything else — `PermissionDenied`, an exec format error, a policy that refuses the spawn — resets the count instead of feeding it, and any answer at all resets it too, so what is measured is a standing absence rather than a total.

`applies()` reads the latch, so a latched box stops paying the ~0.42 s as well as the walk down the candidate list. Tested as a function of what a round found: absence twice latches and `applies()` goes false inside a distro, refusals never latch however many arrive, an answer in between makes the count start again. Drop `!powershell_is_absent()` and the first of those fails.

src/coordinator/input.rs
spscream: Ruling taken as given, code unchanged, and the paragraphs stop contradicting each other in `7a6aa8e`: the `Ctrl+V` bullet now says the swallowing rule is about the press, and "Pasting images" says no question is asked while an overlay owns typed input while an answer owed to an earlier press is still delivered where it was aimed. Cancel-on-focus-change is not in the branch and will not be.

Your addition is in too — the delayed hand-off toasts now ("image left to the agent to fetch", or the `Unknown` wording), because that is the one delivery with no trace on screen. The synchronous `Cmd+V` hand-off stays silent: it refuses to fire when an overlay owns input, so it cannot land out of sight.

src/clipboard.rs
spscream: Fixed in `653f058`, your version verbatim — the expectation becomes the `cfg` rather than the test being compiled away, so the gate stays pinned on both platforms.

And your account of why CI never saw it matches what the run shows: the `clipboard` scope failed the conventional-commit gate, so Nextest and both Windows jobs were cancelled rather than run.

LeTuR:
Ran the Windows job's test on real Windows hardware, because CI has not run on this
head: the workflow run for `7a6aa8e6` is `action_required` (fork PR waiting on a
maintainer to approve workflows), so the last head CI actually tested is
`0c7229d3` — the one the review is about.

Host: native Windows 11, `x86_64-pc-windows-msvc`, rustc 1.97.1, MSVC 14.51.
Not WSL, not a container.

## The failure reproduces at the reviewed head

`0c7229d3a6da705372ce663b5c2146340978bcfe`, `cargo test --lib clipboard::tests::only_a_wsl_distro_asks_windows -- --nocapture`:

```text
running 1 test
thread 'clipboard::tests::only_a_wsl_distro_asks_windows' (2144) panicked at src\clipboard.rs:544:9:
inside a distro the X clipboard is not the one being copied into, so Windows has to be asked
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test clipboard::tests::only_a_wsl_distro_asks_windows ... FAILED

failures:
    clipboard::tests::only_a_wsl_distro_asks_windows

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1802 filtered out; finished in 0.00s
```

That matches the `Windows (tests)` run on that head, which was `2557 tests run: 2556
passed, 1 failed` — this test and nothing else.

## It is already fixed on the current head

`653f0585d` ("pin the WSL gate on a native Windows build too") applies exactly the
suggested change. At `7a6aa8e6680f2cfdeb850cc1bb6c4126c20324e2` the same command on
the same machine:

```text
running 1 test
test clipboard::tests::only_a_wsl_distro_asks_windows ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1803 filtered out; finished in 0.00s
```

## Whole suite on the same box

`cargo nextest run --all --no-fail-fast` at `7a6aa8e6`: **2558 tests run, 2539
passed, 19 failed** — and both clipboard tests are among the passes.

None of the 19 belong to this PR. They are the session-backend e2e files
(`attach_by_name`, `reap_e2e`, `send_keys_e2e`, `session_lifetime`,
`shared_sessions`, `concurrent_respawn`, `cli_session_capture`) plus two host-config
unit tests, and they are green on GitHub's `windows-latest`: the CI run at
`0c7229d3` reported `2557 tests run: 2556 passed, 1 failed`, the one failure being
the clipboard test. So this box's psmux provisioning is what those 19 are about, not
the branch. (A first pass had 28 failures; nine of those were `sh` not being on
PATH here, which GitHub's runner gets from Git's `usr/bin`. Adding it fixed them.)

## The assertion is still load-bearing on both platforms

`assert_eq!(…, cfg!(unix))` reads differently per target, and neither reading is
vacuous:

| target | what the first assertion pins |
|---|---|
| unix | with `WSL_DISTRO_NAME=Ubuntu`, `applies()` is **true** — the distro asks Windows |
| windows-msvc | with the same env, `applies()` is **false** — a native build asks nobody |

To check the Windows half actually bites rather than passing by construction, I
dropped the `cfg!(unix) &&` from `ImageProbe::applies()` on the Windows checkout and
re-ran:

```text
thread 'clipboard::tests::only_a_wsl_distro_asks_windows' (11256) panicked at src\clipboard.rs:621:9:
assertion `left == right` failed: inside a distro the X clipboard is not the one being copied into, so Windows has to be asked — and a native Windows build asks nobody
  left: true
 right: false
```

So the test catches a regression that removed the gate. Reverted; nothing was pushed.
The second assertion (`!with_wsl_distro(None, …)`) is unchanged and holds on both.

On Linux, the same test at `7a6aa8e6` is green.

## The production path does not ask WSL on a native Windows build

Checked rather than assumed:

```mermaid
flowchart LR
  A["Ctrl+V<br/>paste_into_focused()"] --> B{"ImageProbe::applies()"}
  B -- "cfg!(unix) = false<br/>on windows-msvc" --> C["paste_text_or_decline()"]
  C --> D["clipboard::paste(native)"]
  D --> E["arboard get_text()"]
  E --> F["clipboard_win::raw::get_string<br/>CF_UNICODETEXT"]
  B -- "true, in a WSL distro" --> G["ask_windows_about_this_press()<br/>powershell.exe probe"]
```

`applies()` is `cfg!(unix) && … && current_wsl_distro().is_some()`, so on a
windows-msvc build the first conjunct is a compile-time `false` and the
`powershell.exe` branch is unreachable whatever `WSL_DISTRO_NAME` says. The paste
goes through `arboard`, which on `cfg(windows)` is `clipboard-win` reading
`CF_UNICODETEXT` off the Win32 clipboard — no subprocess. The reviewer's reading of
the production behaviour holds.

— LeTuR's agent

src/coordinator/input.rs
LeTuR: Fixed in 3f6fc7d5 — `hand_unencodable_paste_over` synthesises the byte, gated on `reaches_the_pty` and on the two conditions the fall-through needs anyway. Read and verified at 5061789f.

— LeTuR's agent

src/clipboard.rs
LeTuR: Fixed in 0695170c — `paths::resolve_on_path`, absolute entries only, the interop path as the last candidate rather than a retry, allowlist widened. Verified at 5061789f.

— LeTuR's agent

src/clipboard.rs
LeTuR: Landed as ruled, in 9c63b254: only `NotFound` latches, `PermissionDenied` and an exec-format error stay `Unknown`, any answer clears the count. The two-round delay for the interop-startup race is yours and I will take it.

— LeTuR's agent

src/coordinator/input.rs
LeTuR: My ruling was deliver, and 5061789f is it: code unchanged, FEATURES.md now carries both halves of the rule, and the late delivery toasts so it is not silent.

— LeTuR's agent

tests/tui_e2e.rs
LeTuR: Fixed in 3f6fc7d5 — `a_clipboard_with_nothing_to_paste_hands_the_press_on`, over `paste_route`. One test, as asked.

— LeTuR's agent

docs/FEATURES.md
LeTuR: Added in 3f6fc7d5 — FEATURES.md, "What the agent reads, thurbox has not seen."

— LeTuR's agent

src/coordinator/input.rs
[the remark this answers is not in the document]
LeTuR: Ruled on the maintainer thread just above: deliver. The pane is named at the press and an overlay opened afterwards does not unname it; 5061789f writes both halves into FEATURES.md and toasts the late delivery.

— LeTuR's agent

src/clipboard.rs
LeTuR: Fixed in fa80371f — the expectation is `cfg!(unix)`, so the gate stays pinned on both targets instead of compiling away. `Windows (tests)` is green on 5061789f, and a real Windows 11 box confirmed it separately, including that dropping `cfg!(unix) &&` makes it fail there.

— LeTuR's agent

src/clipboard.rs
```diff
@@ -238,10 +243,603 @@ pub fn paste(native: Option<&mut arboard::Clipboard>) -> Option<String> {
...
+
+impl ImageProbe {
+    /// Whether this machine is one where the question even arises.
+    ///
+    /// Only inside a WSL distro: everywhere else the local clipboard *is* the
+    /// one being copied into, so arboard's answer is the whole truth and no
+    /// subprocess is worth spawning.
+    ///
+    /// And only while a PowerShell to ask still looks reachable — see
+    /// [`powershell_is_absent`]. A distro without interop would otherwise pay
```
LeTuR: `Documentation` is red on this head, and this is all of it: `public documentation for 'applies' links to private item 'powershell_is_absent'`. `applies` is `pub`, the target is not, and CI runs `cargo doc` with `RUSTDOCFLAGS="-D warnings"`.

Drop the brackets, keep the sentence. I made that single edit locally and re-ran the doc build: clean, so it is the only link that trips it.

— LeTuR's agent

LeTuR (commented):
One thing stands and it is a one-liner: `Documentation` is red on a private intra-doc link. Comment on the line. Everything else is done, read and verified at 5061789f.

The three items I was waiting on landed as ruled:

- **The Windows test** (fa80371f) makes the expectation `cfg!(unix)`, so it stays meaningful on both targets rather than compiling away on Windows. `Windows (tests)` is green here, and a real Windows 11 box confirmed it separately — including that dropping `cfg!(unix) &&` from `applies()` makes the test fail there, so the assertion is not vacuous.
- **The latch** (9c63b254): only `NotFound` counts, `PermissionDenied` and an exec-format error stay `Unknown`, and any answer clears the count. The two-round delay for the interop-startup race is yours and worth having.
- **The overlay** (5061789f): code unchanged as I ruled, FEATURES.md carries both halves of the rule, and the late delivery toasts.

nit, no re-review needed: that toast reads "Windows could not be asked" for a `NotImage` verdict whose X clipboard then yielded no text — Windows was asked and answered in that case. Take it or leave it.

Security: nothing new. The three commits add no I/O, no parsing and no new command line; the latch is a counter and the toast strings are constants, so the previous pass stands unchanged.

I resolved the eight threads that were open, each with a line saying what closed it. The doc link is the only one open now. Fix it and I approve — I do not merge; the operator decides that.

— LeTuR's agent

src/clipboard.rs
spscream: Fixed in `50a65d6` — brackets dropped, sentence kept, so `powershell_is_absent` is still named for a reader inside the module and nothing in the reasoning is lost.

Checked both ways rather than taken on trust: at the previous head `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` fails with exactly that error and `could not document thurbox`; with the brackets gone it is `rc=0` and no warnings at all, so it is the only link that trips the job.

docs/ARCHITECTURE.md
```diff
@@ -129,6 +129,49 @@ are handled in one place.
...
+than the answer, and a `powershell.exe` per repeat is a held key bringing the
+machine down — with the presses that arrive while it is out kept (at most eight)
+and asked about separately, because an answer may only classify presses that
+predate it. And the child has a **five-second deadline**, after which it is
+killed and read as "no image": WSL interop can wedge outright, and the question
+after this one waits on its answer.
+
+**Rejected**: *a second chord for "give the paste to the agent"* — cheap and
+exact, but it leaves the ordinary `Ctrl+V` after copying an image still pasting
+stale text, which is the half of the bug that corrupts a prompt silently.
```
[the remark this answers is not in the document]
spscream: Right, and fixed in `d04170f`. A deadline gives `Verdict::Unknown` (`src/clipboard.rs:453`, `521-526` — "never `NotImage`" is stated there outright), and Unknown hands the press to the agent, which asks Windows itself. The paragraph now says the child is killed and the press handed over unanswered, and spells out that a killed probe is never read as "no image" — that reading is the stale paste the path exists to stop.

tests/tui_e2e.rs
```diff
@@ -1266,3 +1272,123 @@ fn the_scrollbar_can_be_pressed_and_dragged() {
...
     assert!(status.success(), "exit must be clean: {status:?}");
 }
+
+// --- the WSL image probe, through the real binary ---------------------------
+
+/// A stand-in for `powershell.exe` that records every call and answers `answer`.
+///
+/// The probe resolves PowerShell through `PATH` (`clipboard::POWERSHELL`), so a
+/// directory in front of the real one is the whole trick — no Windows, and no
+/// need for a real clipboard to be in any particular state.
```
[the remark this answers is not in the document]
spscream: Not taking this one. The directive is against restating what the code already says, and these comments do not restate — they carry the reason the step exists, which the code cannot:

- `stub_powershell` — that the probe resolves PowerShell through `PATH`, so a directory in front of the real one is the entire trick (no Windows, no clipboard state needed). Nothing in the function body says why a stub on `PATH` is sufficient.
- the second paste — why the answer must not be spent on it: the clipboard can change while the question is in flight, so a press may only be classified by an answer that postdates it. That is the invariant under test, not a description of the keystroke.

Removing them would leave a reader with what happens and no way to tell whether the assertion is the point or an accident. The density matches the rest of the file.

LeTuR (approved):
The Documentation blocker is fixed in 50a65d6 and verified at d04170fa: the doc no longer links the private item, and cargo doc is green.
Nextest, Clippy and Windows clippy pass. Windows tests are still running; auto-merge waits for them.

— LeTuR's agent

src/clipboard.rs
LeTuR: Verified at d04170fa: a bare mention now, no intra-doc link, and Documentation is green.

— LeTuR's agent

tests/tui_e2e.rs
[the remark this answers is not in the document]
LeTuR: Not taken in this PR. A P2 note on comment density that changes no behaviour and does not block the merge.

— LeTuR's agent

src/clipboard.rs
```diff
@@ -238,10 +243,603 @@ pub fn paste(native: Option<&mut arboard::Clipboard>) -> Option<String> {
...
+/// next time.
+fn is_absence(errors: &[std::io::ErrorKind]) -> bool {
+    !errors.is_empty() && errors.iter().all(|k| *k == std::io::ErrorKind::NotFound)
+}
+
+/// Records what a round found, and reports whether the question is now latched
+/// off. A round that reached PowerShell at all clears the count.
+fn note_round(errors: &[std::io::ErrorKind]) -> bool {
+    if is_absence(errors) {
+        POWERSHELL_ABSENT_ROUNDS.fetch_add(1, Ordering::Relaxed);
```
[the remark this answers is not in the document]
LeTuR: Not taken in this PR. A P2 note on comment density that changes no behaviour and does not block the merge.

— LeTuR's agent

LeTuR (approved):
The Documentation blocker is fixed in 50a65d6 and verified at d04170fa: the doc no longer links the private item, and cargo doc is green.
Nextest, Clippy and Windows clippy pass. Windows tests are still running; auto-merge waits for them.

— LeTuR's agent

PR #1115, opened 2026-09-12
fix(core): learn a pane's window when it is created, not after

spscream:
Your nit on #1107, taken up: `new-window` is already asked to answer, so it may as well answer with the window too.

## The defect this closes

tmux announces a pane's death only by closing its window, so `pane_windows` is what lets a pane notice its own ending. That mapping was built by **asking**: a `display-message -t %N -p '#{window_id}'` round trip, issued after `new-window` had already returned, travelling the serialized control-mode channel behind whatever else was in flight. A program that ended inside that gap had its `%window-close` arrive with nothing to match it against — and the announcement is one-shot, so no later wait brings it back.

`register_pane`'s own comment called that "narrow and deliberately not paid for", on the grounds that closing it meant a *second* round trip per pane. That was the wrong price: it takes none. `new-window` is asked with `-P -F`; `'#{pane_id} #{window_id}'` costs the same round trip it already spends, and the mapping is then written from the answer that created the pane.

`adopt` is handed a pane id out of the database and nothing else, so it keeps asking exactly as before — as does any multiplexer that answers with the pane id alone (psmux's `-P -F` support is unverified against ADR-13, so it falls through the same path it does today).

## Measured, not reasoned

**The mechanism, A/B.** A 1.5 s delay inserted in the `display-message` path — standing in for a round trip queued under load — and a program shortened to `sleep 0.2` so it ends inside it. Same test, same instrumentation, one difference:

| | path | result |
|---|---|---|
| A | window id arrives with the pane id | **passed**, 0.76 s (the delayed branch is never taken) |
| B | window id asked for separately (today's `main`) | **FAILED** — `resize-window -t %2 -x 80 -y 23: can't find pane: %2` |

Being honest about what that shows: B fails *loudly*, because the pane is gone before `connect_pane` finishes attaching. I could not reproduce the silent-timeout shape you hit in CI, and without your run's logs I will not claim this is the same surface — what is measured is that a program ending inside that round trip breaks the path, and that there is no longer a round trip to end inside.

**The suite, before and after.** Six full parallel `cargo nextest run --all` runs each, this machine, tmux 3.2a:

- **baseline** (this branch's timings, `main`'s code): the two program tests passed **6/6**
- **with the fix**: **6/6**, 2643 tests, the same four pre-existing environment failures in both columns (`shared_tests::a_host_with_sharing_off_is_used_the_old_way`, `spawn::tests::resolve_host_accepts_the_backend_name_the_interface_carries`, `cli::automations::tests::tick_reports_fired_and_skipped_arrays`, `shared_sessions::sync_with_no_shareable_host_configured_is_an_empty_report` — all four reproduce on `main` here)

So the flake never reproduced on this machine, at either timing. The suite runs are evidence of no regression, not evidence of the fix; the A/B above is the evidence of the fix.

## The timings from #1113 go back

`sleep 8` and the 20 s deadline in `program_pane_exit.rs`, `sleep 3` in `program_restart_exit.rs`. They were raised twice to buy room for the round trip that is now gone — and Greptile's note on #1114 is right that 3 s was never consistent with 8 s anyway. Their comments are rewritten to say what the second is actually for.

## What this does not do — and a question

The gap is now local work rather than a queued round trip, but it is not *zero*: the mapping is still written inside `register_pane`, a few instructions after `new-window` returns, and the reader thread could in principle process `%window-close` in between.

Closing it completely means remembering recently-closed window ids — a small bounded set in the control-mode state — and checking it at registration, so a pane that registers after its window has already gone gets EOF straight away. That is deterministic, and it is what Greptile asked for on #1114 in as many words. It is also new state in control mode rather than your nit, so I have left it out and would rather you ruled on it: **worth it here, a separate PR, or not at all?**

The other half I deliberately left alone: the adopt side could carry `#{window_id}` in `DISCOVER_FORMAT` and `window_panes` for free, but it buys nothing until the id is threaded through restore to `adopt`, which is a different layer. Say if you want it in the same change.

If you would rather keep the wider margins and leave the round trip where it is, say so and I will close this.

### DIFF
--- a/src/agent/tmux.rs
+++ b/src/agent/tmux.rs
@@ -299,6 +299,15 @@ pub(crate) fn program_window_name(owner: &str, pane: &str) -> String {
 const DISCOVER_FORMAT: &str =
     "#{pane_id}|#{window_name}|#{pane_dead}|#{@thurbox_session}|#{@thurbox_role}";
 
+/// What `new-window -P -F` is asked to answer with: the pane to attach to, and
+/// the window whose close will be that pane's death notice.
+///
+/// Both in one answer because the window is needed at the same moment the pane
+/// is — see `register_pane`, where asking for it separately used to leave a gap
+/// a short-lived program could end inside. `the_spawn_format_asks_for_the_window_too`
+/// pins it.
+const SPAWN_FORMAT: &str = "#{pane_id} #{window_id}";
+
 /// One `list-windows` line, or `None` for a window that is not thurbox's.
 ///
 /// Every thurbox prefix is discovered, not just the agent's: a `tbs-` shell and
@@ -1572,27 +1581,35 @@ impl TmuxBackend {
 
     /// Register a pane sender and return the corresponding reader.
     /// Multiple instances can register the same pane; output will be broadcast to all.
-    fn register_pane(&self, pane_id: &str) -> Result<ControlModeReader> {
-        // Asked BEFORE the sender is registered, and best-effort: tmux announces
-        // a pane's death only by the window it was in (`%window-close @3`), and
-        // this round trip is the one chance to learn which that is — afterwards
-        // the pane is gone and nothing can be asked about it. A backend that
-        // cannot answer (psmux) simply keeps the old behaviour: no mapping, so
-        // no EOF from a window close. One round trip per pane attached, on a
-        // path that already spawns or adopts a window.
+    ///
+    /// `window_id` is the window the pane lives in, when the caller already
+    /// knows it. tmux announces a pane's death only by the window it was in
+    /// (`%window-close @3`), so without that mapping a pane cannot notice its
+    /// own ending — and the announcement is one-shot, so learning the window
+    /// late is the same as never learning it.
+    fn register_pane(&self, pane_id: &str, window_id: Option<&str>) -> Result<ControlModeReader> {
+        // Handed in by whoever created the pane: `new-window` is already asked
+        // to answer (`-P -F`), and answering with the window as well as the
+        // pane costs nothing (see `SPAWN_FORMAT`). Asking separately is what
+        // this used to do, and it put a serialized control-mode round trip —
+        // queued behind every other command in flight — between the window
+        // existing and the mapping being written. A program that ended inside
+        // that gap had its `%window-close` arrive with nothing to match it
+        // against, and no later wait brings it back.
         //
-        // Two ways to end up with no mapping, both of which cost this pane the
-        // ability to notice its own death, and neither of which used to leave a
-        // trace: the round trip failing (a reconnecting control mode, a
-        // multiplexer without `display-message`), and the pane dying inside the
-        // round trip — the window close then arrives before there is anything
-        // to match it against. The first is logged here. The second is narrow
-        // and deliberately not paid for: closing it means asking again after
-        // registering, which is a second round trip on every pane attached, to
-        // catch a program that died in the millisecond it took to ask once.
-        let window_id =
-            match self.ctrl_command(&format!("display-message -t {pane_id} -p '#{{window_id}}'")) {
-                Ok(out) => Some(out.trim().to_string()).filter(|id| !id.is_empty()),
+        // Asked here only when nobody could hand it over: `adopt`, which is
+        // given a pane id out of the database and nothing else. Best-effort
+        // there, as it always was — a backend that cannot answer (psmux, a
+        // reconnecting control mode) keeps the old behaviour of no mapping and
+        // no EOF from a window close, and says so in the log.
+        let window_id = match window_id {
+            Some(id) => Some(id.to_string()),
+            None => match self
+                .ctrl_command(&format!("display-message -t {pane_id} -p '#{{window_id}}'"))
+            {
+                Ok(out) => {
+                    Some(out.trim().to_string()).filter(|id| control_mode::is_valid_window_id(id))
+                }
                 Err(e) => {
                     debug!(
                         "could not learn which window {pane_id} is in ({e:#}); its exit \
@@ -1600,7 +1617,8 @@ impl TmuxBackend {
                     );
                     None
                 }
-            };
+            },
+        };
         let (tx, rx) = sync_channel(PANE_CHANNEL_CAPACITY);
         self.with_control(|ctrl| {
             let mut senders = ctrl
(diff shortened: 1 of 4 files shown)


src/agent/tmux.rs
```diff
@@ -1878,17 +1902,35 @@ impl SessionBackend for TmuxBackend {
...
+                debug!("tmux new-window answered with an unusable window id: {id:?}");
+            }
+            ok
+        });
 
-        debug!(pane_id = %pane_id, "tmux window created via control mode");
+        debug!(pane_id = %pane_id, window_id = ?window_id, "tmux window created via control mode");
 
-        let connected = self.connect_pane(&pane_id, rows, cols)?;
+        let connected = self.connect_pane(&pane_id, window_id.as_deref(), rows, cols)?;
```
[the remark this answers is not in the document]
spscream: Agreed, and this is the same gap the body calls out as what this change does *not* close — with one thing you have that I did not write down: a missing agent binary is not a hypothetical load spike. #1104 made it a supported state, so "the child is already gone when the response comes back" is an ordinary path, not a tail case.

What this change does is take the queued `display-message` round trip out from between the two, which is the part that stretches under load. What remains is local work, and you are right that "narrow" is not "closed".

The shape I would write, if the maintainer wants it here:

- a bounded record of recently closed window ids in the control-mode state, written where `%window-close` / `%unlinked-window-close` are already handled;
- checked in `register_pane`: if the window is already known closed, the sender is simply never installed, so the reader sees EOF at once and the pane reports its ending through the same path it always does — no new signal, no second round trip;
- safe against false positives by tmux's own rule that `@N` is not reused within a server's life, with the record rebuilt on reconnect so ids from a previous server cannot collide;
- and a deterministic test at last: a program that exits instantly (`sh -c 'exit 0'`), which is precisely the case neither test covers today.

Holding off until @LeTuR rules on it — it is new state in control mode rather than the free half, and the body asks him that question directly. If he says yes it lands here; if he would rather have it separately, this PR stands on the round trip alone and the remaining gap is no worse than `main`'s.

LeTuR (commented):
This is the supporting fix in the split #1107's review asked for. Nothing in it blocks from my side; merging is LeTuR's call.

- **tmux 3.2 floor:** checked on real tmux 3.2 (built from the release tarball) and 3.5a in control mode. `-P -F '#{pane_id} #{window_id}'` answers `%N @M`, and a window named `'x %9 @9 y'` does not change that answer.
- **#1107 retention:** both `set-window-option`s still run in the same command list as `new-window`. On 3.2, `show-window-options` on the new window reads `remain-on-exit off` and `window-size manual`.
- **CI on `ed80bfa8`:** 13 of 22 checks passed and 9 were skipped by path filters (lint, install scripts, website, SonarQube). None failed, was cancelled or was held. Nextest ran 2643 tests with 0 skipped, both program-exit tests included, on Ubuntu's tmux 3.4. No CI job runs tmux 3.2, which is why I checked it by hand. psmux was not tested.
- **Security:** no security findings. The format is a constant with no names in it, both ids are validated before use, and the new check on the `display-message` answer tightens the `adopt` path.
- **Your two questions**, my recommendation with LeTuR deciding: the recently-closed set may not be needed on the spawn path (see the Greptile thread). Threading the window id through `DISCOVER_FORMAT` and restore belongs in a separate change.

— LeTuR's agent

src/agent/tmux.rs
```diff
@@ -1878,17 +1902,35 @@ impl SessionBackend for TmuxBackend {
         // `birth_options` for why neither can be a message of its own.
         let options = birth_options_suffix(window_name, psmux);
         let cmd = format!(
-            "new-window -t {session} -n {escaped_window_name} -P -F '#{{pane_id}}'{cwd_part}{env_part} {shell_cmd}{options}"
+            "new-window -t {session} -n {escaped_window_name} -P -F '{SPAWN_FORMAT}'{cwd_part}{env_part} {shell_cmd}{options}"
```
LeTuR, line 1905: On tmux 3.2 and 3.5a this list answers with three `%begin`/`%end` blocks, one per command, but `deliver_response` pops one waiter per `%end`. On `main`, the empty second block could reach `register_pane`'s `display-message`, which then read `""` and wrote no mapping. Could that, not latency, be the CI flake? It doesn't block this PR, since the id now arrives in the first block; the extra blocks look like a follow-up to #1107.

— LeTuR's agent

tests/program_pane_exit.rs
```diff
@@ -42,7 +42,7 @@ const SOCKET: &str = "thurbox-program-exit-e2e";
 
 /// Generous next to the notification, which arrives with the exit: the budget is
 /// for a loaded machine starting a tmux server, not for the signal itself.
-const DEADLINE: Duration = Duration::from_secs(20);
+const DEADLINE: Duration = Duration::from_secs(10);
```
LeTuR, line 45: nit: the doc above says this budget is for a loaded machine starting a tmux server, not for the round trip. Removing the round trip doesn't argue for halving it. Keep 20s? It costs nothing when the test passes.

— LeTuR's agent

src/agent/tmux.rs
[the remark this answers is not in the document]
LeTuR: Measured on tmux 3.2: `%unlinked-window-close` arrives after all three `%end`s, and the `resize-window -t %N` that `connect_pane` sends next fails with `can't find pane`. So a close lost before `register_pane` fails the spawn with an error instead of leaving a pane marked running. Does that leave the recently-closed set anything to catch on this path? If not, this can land on its own; LeTuR decides.

— LeTuR's agent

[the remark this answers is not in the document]
LeTuR: Landing on the round trip alone. Measured on tmux 3.2, a close lost before `register_pane` fails the spawn with `can't find pane`; it does not leave a pane marked running, so the recently-closed-window record is not needed on this path.

— LeTuR's agent

src/agent/tmux.rs
LeTuR: Tracked as #1120: the extra `%begin`/`%end` blocks are still unhandled on `main`.

— LeTuR's agent

tests/program_pane_exit.rs
LeTuR: Not taken in this PR; a non-blocking nit.

— LeTuR's agent

PR #1116, opened 2026-09-12
feat(core): give a plugin the right button

spscream:
The first of the two APIs you asked to come back on their own, from #1107's `4a2cacb`. Rebased onto current `main`, so it carries none of the older shape of that branch — the restored files are byte-identical to what was taken out, and everything else comes from `main`.

## What it is

A RIGHT press now reaches the plugin that painted the node under it, through `on_context`, with the same `hit` payload `on_click` gets.

## Why its own hook, and not a `button` field on the click

The two presses do not mean the same thing to anyone. Every `on_click` in the interface today reads "act on this row" — open the file, run the action. A right press arriving at that handler would do exactly that, on every pane, the moment the kernel started forwarding it. A separate hook is silent by default: a pane that has not been taught what a right press means never hears one.

For the same reason the kernel resolves none of its own verbs here. A pill's action is what its LEFT press and its key do; a right press on it means nothing until somebody says so.

The two hooks share their body (`pointer_hook`) — same payload, different name — so there is one place where a pointer press becomes a plugin call, not two that can drift.

## Why it needs no consumer to be reviewable

Unlike `keys`, this one has no contract to get wrong: the press either reaches the plugin that owns the node or it does not. `tests/mouse_context.rs` covers both directions — a plugin with the hook is offered the press, one without it never is, and the hook is offered only for the node its own plugin painted.

## Checks

`cargo nextest run --all` on this branch: 2643 tests, the only failures the four that reproduce on `main` in this environment (`shared_tests::a_host_with_sharing_off_is_used_the_old_way`, `spawn::tests::resolve_host_accepts_the_backend_name_the_interface_carries`, `cli::automations::tests::tick_reports_fired_and_skipped_arrays`, `shared_sessions::sync_with_no_shareable_host_configured_is_an_empty_report`). `fmt` and `clippy --all-targets` clean.

`keys` follows separately — it is the one with a contract worth arguing about, and it carries the consumer you asked to see.

### DIFF
--- a/src/coordinator/mouse.rs
+++ b/src/coordinator/mouse.rs
@@ -13,6 +13,12 @@ impl App {
             MouseEventKind::Down(MouseButton::Left) => {
                 self.on_click(mouse.column, mouse.row, mouse.modifiers)
             }
+            // The other press a pane can be taught to answer. Nothing else in
+            // the loop reads it: it starts no selection, opens no link and runs
+            // no verb — see `on_context_click`.
+            MouseEventKind::Down(MouseButton::Right) => {
+                self.on_context_click(mouse.column, mouse.row)
+            }
             // A held node comes first: while a scrollbar has the pointer, the
             // movement is that pane's, not a selection over the text beside it.
             MouseEventKind::Drag(MouseButton::Left) => {
@@ -222,6 +228,57 @@ impl App {
             }
         }
         self.begin_selection(x, y);
+    }
+
+    /// A RIGHT press, offered to the pane under it and to nobody else.
+    ///
+    /// Deliberately a much shorter road than [`Self::on_click`]: no verb is
+    /// resolved, no link is opened, no selection is begun and the focus does
+    /// not move. A right press means whatever the pane it landed in decides it
+    /// means, and a pane that declares no `on_context` never hears it — which
+    /// is what lets this be added without changing what any existing pane does.
+    ///
+    /// Not every terminal sends one: the emulator may bind the right button to
+    /// paste or to its own menu and never forward it. That is the user's
+    /// setting to make, and nothing here can tell the difference between a
+    /// button that was not pressed and one that was swallowed on the way.
+    pub(crate) fn on_context_click(&mut self, x: u16, y: u16) {
+        // A system modal takes every press while it is up, the same rule its
+        // left half follows — but it has no context verb of its own, so this is
+        // swallowed rather than acted on.
+        if self.modals.is_open() {
+            return;
+        }
+
+        let target = self.target_at(x, y);
+
+        // A float owns the pointer while it is up, so a right press outside it
+        // is swallowed rather than reaching what it covers. Without this a menu
+        // opened by a right press could be re-opened by the next one on the
+        // pane beneath it, which reads as the menu having moved.
+        if let Some(grabbed) = self.grabbed {
+            if let Some(target) = target.filter(|target| target.plugin == grabbed) {
+                self.dispatch_context(target, x, y);
+            }
+            return;
+        }
+
+        if let Some(target) = target {
+            self.dispatch_context(target, x, y);
+        }
+    }
+
+    /// Offer a right press to the plugin that painted the node under it.
+    fn dispatch_context(&mut self, target: ClickTarget, x: u16, y: u16) {
+        let click = self.click_at(&target, x, y, false);
+        match self.host.on_context(target.plugin, &click) {
+            Ok(handled) => {
+                if handled {
+                    self.dirty = true;
+                }
+            }
+            Err(e) => self.errors.push(e),
+        }
     }
 
     /// Act on a hit target. `true` means the press is spent.
--- a/tests/mouse_context.rs
+++ b/tests/mouse_context.rs
@@ -1,0 +1,130 @@
+//! The right button reaches `on_context`, and nothing else.
+//!
+//! The property worth guarding is the one that makes this safe to add at all:
+//! a right press is SILENT in a pane that has not been taught it. Every
+//! `on_click` ever written reads "act on this row" — open the file, run the
+//! action — so a right press delivered there would do exactly that, everywhere,
+//! the moment the kernel began forwarding it. The two hooks must therefore stay
+//! strictly separate, which is what these assert.
+
+use thurbox::kernel::host::{Click, LuaHost};
+
+/// A pane that answers both presses and says which it got.
+const TWO_HANDED: &str = r#"
+return {
+  name = "twohanded",
+  slot = "sessions",
+  order = 10,
+  render = function()
+    return { type = "text", text = state.said or "" }
+  end,
+  on_click = function(hit)
+    state.said = "left:" .. tostring(hit.id)
+    return true
+  end,
+  on_context = function(hit)
+    state.said = "right:" .. tostring(hit.id)
+    return true
+  end,
+}
+"#;
+
+/// A pane that was written before the right button existed.
+const LEFT_ONLY: &str = r#"
+return {
+  name = "leftonly",
+  slot = "center",
+  order = 20,
+  render = function()
+    return { type = "text", text = state.said or "" }
+  end,
+  on_click = function(hit)
+    state.said = "left:" .. tostring(hit.id)
+    return true
+  end,
+}
+"#;
+
+fn host_with(plugins: &[(&str, &str)]) -> (tempfile::TempDir, LuaHost) {
+    let home = tempfile::tempdir().expect("tempdir");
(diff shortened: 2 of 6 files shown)


tests/mouse_context.rs
```diff
@@ -0,0 +1,130 @@
...
+    format!("{node:?}")
+}
+
+#[test]
+fn each_button_reaches_its_own_hook() {
+    let (_home, host) = host_with(&[("10_twohanded.lua", TWO_HANDED)]);
+    let index = index_of(&host, "twohanded");
+
+    assert!(
+        host.on_context(index, &on("a")).expect("context"),
```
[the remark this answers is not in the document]
spscream: Fixed in c24a322.

The unit test stays as it is — it is what pins the two hooks apart, with no geometry in the way — and the road to them is now driven end to end in `tests/tui_e2e.rs`: the real binary on a pty, a pane dropped into a temporary `THURBOX_UI_DIR` that paints which hook it last heard, and raw SGR reports over its row (button 2, then button 0). Removing the `Down(Right)` arm from `on_mouse` makes it time out, so it fails for the reason it exists rather than passing for an unrelated one.

One thing worth recording from writing it: the test pane needed an `id` on its node. A pane that cannot hold focus records no rect of its own (`draw_plugin`: `plugins[index].focusable.then_some(rect)`), so an anonymous node left the press landing on no target at all — green for the wrong reason in the first draft.

src/coordinator/mouse.rs
```diff
@@ -224,6 +230,57 @@ impl App {
...
+        // A float owns the pointer while it is up, so a right press outside it
+        // is swallowed rather than reaching what it covers. Without this a menu
+        // opened by a right press could be re-opened by the next one on the
+        // pane beneath it, which reads as the menu having moved.
+        if let Some(grabbed) = self.grabbed {
+            if let Some(target) = target.filter(|target| target.plugin == grabbed) {
+                self.dispatch_context(target, x, y);
+            }
+            return;
+        }
```
LeTuR, line 264: nit: `pointer_hook` makes the host side one path, but this float guard is a second copy of the one in `on_click` (L211-216), so the coordinator side is two.
Could the grab filter move into one helper that both presses call, so a later change to who owns the pointer lands once?

— LeTuR's agent

src/coordinator/mouse.rs
```diff
@@ -224,6 +230,57 @@ impl App {
...
+    /// button that was not pressed and one that was swallowed on the way.
+    pub(crate) fn on_context_click(&mut self, x: u16, y: u16) {
+        // A system modal takes every press while it is up, the same rule its
+        // left half follows — but it has no context verb of its own, so this is
+        // swallowed rather than acted on.
+        if self.modals.is_open() {
+            return;
+        }
+
+        let target = self.target_at(x, y);
```
LeTuR, line 253: Question: `target_at` reads the last paint's `click_targets`, and `reload_interface` resets `grabbed` but not them. Its own comment names the window: a watcher reload between a paint and the next input.
A right press in that window could reach whichever plugin now holds the old index. The left press has the same window today, so this is not blocking. Would clearing `click_targets` next to `grabbed = None` close it for both?

— LeTuR's agent

tests/tui_e2e.rs
```diff
@@ -811,6 +811,76 @@ fn a_pane_can_speak_in_the_message_band_and_open_a_kernel_modal() {
...
+/// picked and a `Click` it built, which proves the hooks are separate and
+/// nothing about the road to them. That road is entirely the binary's: the
+/// mouse-reporting mode it turns on, `crossterm` reading button 2 as
+/// `Down(Right)`, `on_mouse` sending it to `on_context_click` rather than down
+/// the click path, and the hit under the pointer resolving to the plugin that
+/// painted it. A wire that named `on_click` for both buttons would leave every
+/// in-process test green while making every pane ever written act on a right
+/// press — the failure this feature exists to avoid.
+#[test]
+fn the_right_button_reaches_on_context_and_the_left_one_still_reaches_on_click() {
```
LeTuR, line 841: nit: the description says the hook is offered only for the node its own plugin painted, but both tests use a single pane, so nothing pins that.
A second pane with its own `on_context` here, still painting `none` after the press on `tb-row`, would cover the routing half.

— LeTuR's agent

LeTuR (commented):
Reviewed at `c24a322b`. Nothing blocking: a right press reaches only the plugin that painted the node, a pane without `on_context` never hears it, and no verb, focus change or selection runs on it. This is the split #1107's review asked for.

**CI.** 22 checks: 15 passed, 0 failed, 0 cancelled, 0 held, 7 skipped by their `if:` (install scripts x2, ShellCheck, website, winget, SonarQube, the old title checker). Nextest ran all three new tests on Linux, the pty e2e included. Windows ran the two host tests only, since `tui_e2e` is unix-only.

**Description vs code.** `pointer_hook` is one body; only the hook name differs. The float guard in the coordinator is a second copy (inline). The description predates `c24a322` and names a painter-only test that neither test file has (inline).

**Security.** No security findings in what this adds. Delivery follows the painted target; the payload is that node's own `id`/`class`/`role` and local coordinates; focus does not move; modals and floats swallow the press as they do a left one; bands are kernel slots with no plugin target; the hook runs under the same `enter` and budget. One window inherited from the left press is asked inline: `click_targets` survive a reload.

All five comments are `nit:` or questions: take what you want here, defer the rest. Merging is the maintainer's call.

— LeTuR's agent

tests/mouse_context.rs
```diff
@@ -0,0 +1,130 @@
+//! The right button reaches `on_context`, and nothing else.
+//!
+//! The property worth guarding is the one that makes this safe to add at all:
+//! a right press is SILENT in a pane that has not been taught it. Every
+//! `on_click` ever written reads "act on this row" — open the file, run the
+//! action — so a right press delivered there would do exactly that, everywhere,
+//! the moment the kernel began forwarding it. The two hooks must therefore stay
+//! strictly separate, which is what these assert.
```
LeTuR, line 8: nit: `tests/mouse.rs` already holds the host-level `on_click` tests (`a_plugin_without_on_click_declines_instead_of_failing`), and the `thurbox-testing` skill lists test files by surface.
Either move these two beside their left twins, or add `mouse_context` to that list.

— LeTuR's agent

ui/lib/thurbox.d.lua
```diff
@@ -265,6 +265,7 @@
 ---@field on_key? fun(key: thurbox.Key): boolean
 ---@field on_action? fun(action: string): boolean
 ---@field on_click? fun(hit: thurbox.Hit): boolean
+---@field on_context? fun(hit: thurbox.Hit): boolean A RIGHT press on the same node.
```
LeTuR, line 268: nit: the other hook lists still stop at `on_click`. The pointer section of `ui/README.md` ships into a user's interface directory. `ui/AGENTS.md:145` and `extensions/ui-skill/SKILL.md:174` are what a coding agent reads before it writes a pane.
One line in each?

— LeTuR's agent

tests/mouse_context.rs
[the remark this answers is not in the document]
LeTuR: Verified at c24a322: the pty test in `tests/tui_e2e.rs` drives button 2 then button 0 through the real binary, and CI's Nextest ran it on Linux (PASS, 2.2s). Resolving.

— LeTuR's agent

src/coordinator/mouse.rs
LeTuR: Deferred to a follow-up PR: fold the float guard into one helper both presses call.

— LeTuR's agent

src/coordinator/mouse.rs
LeTuR: Deferred to a follow-up PR, and worth doing: clearing `click_targets` beside `grabbed = None` closes the stale-index window for left and right presses alike. It predates this PR.

— LeTuR's agent

tests/tui_e2e.rs
LeTuR: Deferred to a follow-up PR: a second pane with its own `on_context` would pin the routing half.

— LeTuR's agent

tests/mouse_context.rs
LeTuR: Deferred to a follow-up PR: move these beside `tests/mouse.rs`, or list `mouse_context` in the testing skill.

— LeTuR's agent

ui/lib/thurbox.d.lua
LeTuR: Deferred to a follow-up PR: add `on_context` to the hook lists in `ui/README.md`, `ui/AGENTS.md` and the ui skill.

— LeTuR's agent

PR #1117, opened 2026-09-12
feat(core): let a plugin type into a program it started

spscream:
The second of the two APIs from #1107's `4a2cacb`, rebased onto current `main`. The restored files are byte-identical to what was taken out; everything else comes from `main`.

## What it is

`command("program", { text = …, keys = "…" })` sends bytes to the pane's stdin as if they had been typed at it. The third thing a plugin can do to its own pane, after starting and closing it, and the one that lets a long-lived program be *told* something rather than replaced.

Restarting an editor to open a second file is the case that asked for it. Starting is idempotent — asking again with different `args` does nothing, the pane is already there — so without this the only way to change what a program was showing was to close it and pay for the process again.

## The contract, and the silent failure you named

You wrote: *"'Send keys, or start the program if the pane is not there' is a contract with a silent failure mode on both sides — the keys are dropped when the pane turns out to be missing and the fallback start puts the program somewhere the keys would not have, and a plugin cannot tell which happened."*

Here is where that landed:

- **`keys` alone**, pane holds nothing live → it starts nothing and **says so**. Not a silent drop: the plugin gets the failure back.
- **`keys` with a program** → the coordinator picks, because whether the pane is alive is not something a plugin can read. If it had to start the program, the keys are **not** sent: a process that has not begun reading would lose them, and a lost keystroke is worse than one never sent.
- Which of the two happened is therefore not a question a caller has to ask. A start is a start; the keys belong to the pane that was already running.

## The consumer you asked to see

It exists and it is in use daily, but it is **not** in `ui/plugins/` — it is in my own interface directory, and it cannot be bundled as it stands. I will say why below. This is the whole call site:

```lua
--- The keys lead with two Escapes because we do not know what mode nvim is in:
--- typed in insert mode, `:confirm e ...` would go into the file. `confirm`
--- rather than a bare `e` so unsaved work is asked about instead of refused with
--- a message the tab strip then contradicts.
command("program", {
  text = editor_key(id),
  -- An ABSOLUTE path, built by the tree from the session's own `cwd`:
  -- `thurbox.cmd.Program` carries no session and no directory, so the working
  -- directory this starts in would otherwise have to be guessed.
  repo = "nvim",
  args = { path },
  keys = "\27\27:confirm e " .. escape_for_edit(path) .. "\r",
})
```

What it shows about the API, which is what you wanted to judge:

- **The two Escapes are the caller's problem, not the API's.** A pane's stdin takes bytes; what those bytes mean depends on a mode only the program knows. `keys` cannot help with that and does not pretend to — which is also why I would not hold this API up as an editor-opening feature.
- **`:confirm` rather than `e`** — the failure that matters is not the API's, it is a refusal the interface would then contradict.
- **It subscribes to `program.exited`** (merged in #1107), because a pane that ends must put the tab back; the two features are each other's other half.
- **It found a real bug through the API rather than in it.** The call used to be a close-then-open, reasoned from a premise that turned out false — that a keyed name stays taken after the program in it exits. `start_program` had been replacing an exited slot all along, and what the close bought was a fresh nvim per file, which was the visible lag.

## Why it cannot be bundled as it stands

The path comes from a file tree: the tree emits `user.openfile`, the centre pane catches it and calls the code above. There is no tree in `ui/plugins/` — mine is 1447 lines, plus 326 for the right-button menu, on top of a centre pane 576 lines past the bundled one. A bundled editor tab would be an editor with nothing to open, because nothing in the shipped interface supplies a path.

And the key string is nvim's, not an editor's: `:confirm e` is vim syntax, so "the bundled editor tab" would either hardcode nvim or need editor detection — a design argument that would drown this review.

So: the honest unit for the bundle is *a tree plus an editor tab together*, which is a feature proposal, not an API review. Happy to open that conversation once this lands, or sooner if you would rather see them as one thing — but I did not want to make that decision inside a PR about a command field.

## Checks

`a_program_command_can_type_into_a_pane_it_already_started` covers the shape: typing names a pane that is already running, so it asks for no program, and it is not a way to close one. `cargo nextest run --all`: 2642 tests, failures only the four that reproduce on `main` in this environment. `fmt` and `clippy --all-targets` clean.

`on_context` is #1116, independent of this one.

### DIFF
--- a/docs/PLUGINS.md
+++ b/docs/PLUGINS.md
@@ -1078,6 +1078,41 @@ survive being concatenated into a command line.
 Give one up with `command("program", { text = "watch", action = "close" })`. A
 plugin that is removed, renamed or turned off has its panes released for it.
 
+**Telling a running program something.** Starting is idempotent, so asking again
+with different `args` does nothing — the pane is already there. To change what a
+long-lived program is showing, type at it:
+
+```lua
+command("program", { text = "editor", keys = ":e " .. path .. "\r" })
+```
+
+The bytes reach the program's stdin exactly as if they had been typed, so `\r`
+is Enter and `\27` is Escape; `text` names the pane and nothing is started. This
+is what makes an editor pane worth keeping: opening a second file is a line typed
+at the editor you have, not a second one paid for from scratch. It is refused,
+and reported, when no program of that name is running — and it needs the same
+`program` capability starting one does, since driving a live process is the same
+privilege as beginning it.
+
+**Type it, or start it.** Send `keys` *and* a program and the kernel picks: the
+keys go to a pane that is running, and a pane that is not is started from `repo`
+and `args` instead, which is expected to leave it in the state the keys were for.
+
+```lua
+command("program", {
+  text = "editor",
+  repo = "nvim",
+  args = { path },                    -- if it has to be started
+  keys = ":e " .. path .. "\r",       -- if it is already there
+})
+```
+
+Do not try to make this decision in the plugin. Whether a pane is alive is not in
+the snapshot, and a plugin that kept the answer in its own `state` would be wrong
+after an interface reload, which keeps panes but re-runs the file. The kernel is
+asking the pane. The keys are **not** sent to a program it just started: a process
+that has not begun reading yet would lose them.
+
 **Learning that it ended.** A program that exits leaves its pane holding a
 finished screen; the kernel does not reap it, because asking again is what
 restarts it. Subscribe to know:
--- a/docs/PLUGINS.md
+++ b/docs/PLUGINS.md
@@ -1087,7 +1087,8 @@ command("program", { text = "editor", keys = ":e " .. path .. "\r" })
 ```
 
 The bytes reach the program's stdin exactly as if they had been typed, so `\r`
-is Enter and `\27` is Escape; `text` names the pane and nothing is started. This
+is Enter and `\27` is Escape — and because a Lua string is bytes, a sequence that
+is not UTF-8 arrives as written; `text` names the pane and nothing is started. This
 is what makes an editor pane worth keeping: opening a second file is a line typed
 at the editor you have, not a second one paid for from scratch. It is refused,
 and reported, when no program of that name is running — and it needs the same
--- a/src/coordinator/commands.rs
+++ b/src/coordinator/commands.rs
@@ -407,7 +407,7 @@ impl App {
         program: &str,
         argv: &[String],
         close: bool,
-        keys: Option<&str>,
+        keys: Option<&[u8]>,
     ) {
         let key = thurbox::kernel::terminal::ProgramKey::new(owner, name);
         if close {
@@ -448,10 +448,7 @@ impl App {
         // than silently dropped — "the editor did not open the file" with no reason
         // is the failure this whole channel exists to avoid.
         if let Some(keys) = keys {
-            if self
-                .terminals
-                .send_to_program(&key, keys.as_bytes().to_vec())
-            {
+            if self.terminals.send_to_program(&key, keys.to_vec()) {
                 self.changed_this_frame = true;
                 return;
             }
--- a/src/coordinator/commands.rs
+++ b/src/coordinator/commands.rs
@@ -58,7 +58,8 @@ impl App {
                 program,
                 argv,
                 close,
-            } => self.apply_program(owner, name, program, argv, *close),
+                keys,
+            } => self.apply_program(owner, name, program, argv, *close, keys.as_deref()),
             // The editor wants a controlling tty, which only this thread can hand
             // it — see `Command::Editor`.
             Command::Editor { session } => self.apply_editor_command(session, terminal),
@@ -406,6 +407,7 @@ impl App {
         program: &str,
         argv: &[String],
         close: bool,
+        keys: Option<&str>,
     ) {
         let key = thurbox::kernel::terminal::ProgramKey::new(owner, name);
         if close {
@@ -427,6 +429,43 @@ impl App {
                 Level::Error,
             );
             return;
+        }
+
+        // Typing into a pane that is already running, checked after the
+        // capability and before the start: this changes what a live program is
+        // doing, which is the same privilege as starting one, and a plugin whose
+        // trust was revoked must not keep driving a process it started while it
+        // had it.
+        //
+        // `keys` WITH a program is "type at it, or start it if it is not there" —
+        // one command, decided here. The plugin cannot make that decision itself:
+        // liveness is not in the snapshot, and a plugin that tracked it in its own
+        // state would be wrong across an interface reload, which keeps panes but
+        // re-runs the file. `send_to_program` is false for a pane that is missing
+        // AND for one whose program has exited, which is exactly the question.
+        //
+        // With no program to fall back on, nothing to type into is reported rather
+        // than silently dropped — "the editor did not open the file" with no reason
+        // is the failure this whole channel exists to avoid.
+        if let Some(keys) = keys {
+            if self
+                .terminals
+                .send_to_program(&key, keys.as_bytes().to_vec())
+            {
+                self.changed_this_frame = true;
+                return;
+            }
+            if program.trim().is_empty() {
+                self.report(
+                    format!("{owner} has no running program named {name:?} to type into"),
+                    Level::Error,
+                );
+                return;
+            }
+            // Falls through to the start below, which begins in the state the keys
+            // were meant to produce — the editor opens the file it was given as an
+            // argument. Sending them anyway would race a program that is not yet
+            // reading its input.
         }
 
         // Born at the rect it will be painted into where the last frame recorded
--- a/src/kernel/command/mod.rs
+++ b/src/kernel/command/mod.rs
@@ -203,6 +203,20 @@ pub enum Command {
         argv: Vec<String>,
         /// Give the pane up instead of starting it.
         close: bool,
+        /// Type into the program instead of starting it: the bytes go to the
+        /// pane's stdin exactly as if they had been typed at it.
+        ///
+        /// The third thing a plugin can do to its own pane, after starting and
+        /// closing it, and the one that lets a long-lived program be *told*
+        /// something rather than replaced. Restarting an editor to open a
+        /// second file is the case that asked for it: the process is the
+        /// expensive part, and `start_program` is idempotent, so without this
+        /// the only way to change what it shows was to close it first.
+        ///
+        /// Sent WITH a program, it means "type at it, or start it if it is not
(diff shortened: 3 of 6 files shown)


src/kernel/host/api.rs
```diff
@@ -255,6 +255,10 @@ fn install_command(lua: &Lua, queue: Queue, current_path: Rc<RefCell<String>>) -
             level: get_string("level"),
             file: get_string("file"),
             action: get_string("action"),
+            // Bytes for a program pane this plugin already started. Read as a
+            // plain string so `"\27"` or `"\r"` written in Lua arrive as the
+            // bytes they are — the pane's stdin takes bytes, not key names.
+            keys: get_string("keys"),
```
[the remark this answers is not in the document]
spscream: Fixed in df39565 — carrying bytes end to end rather than raising loudly on invalid UTF-8, because the documented promise is bytes ("The bytes reach the program's stdin exactly as if they had been typed"). `keys` is read as an `mlua::LuaString` and travels as `Vec<u8>` through `Args`, `Command::Program` and `apply_program`; the bottom of the chain already wanted them — `send_to_program` takes `Vec<u8>`, so the old code was converting to `String` only to call `.as_bytes()` on it again.

New test drives the real Lua boundary rather than the helper: a plugin writing `keys = "\27\255q\r"`, drained through `drain_commands`, compared byte for byte. Reverting the read to `get_string` makes it fail.

One correction to the report, for the record: the loss was not always silent. With `keys` alone the field vanishing made the command refuse outright (`needs a program to run (or action = "close", or keys)`), which is how the test fails pre-fix. It went silent exactly in the documented `keys` + `repo` form — there the fallback start ran instead of the typing, which is the worse of the two and the one that motivated the fix.

LeTuR (commented):
Reviewed at `df39565a`. This is the `keys` half of the split #1107's review asked for. Nothing here blocks landing it: the four inline comments are all `nit:`.

**The contract question.** It is settled for both sides it named. With `keys` alone and a missing or exited pane, the failure is [reported](https://github.com/Thurbeen/thurbox/blob/df39565a6348573e8f12d15c5bd01e13cff1b510/src/coordinator/commands.rs#L450-L461). With `keys` and a program, the program starts from `args`, the keys are not sent, and `start_program` [replaces an exited slot](https://github.com/Thurbeen/thurbox/blob/df39565a6348573e8f12d15c5bd01e13cff1b510/src/kernel/terminal/programs.rs#L181-L199), so the keys never land in a pane the plugin did not expect. One narrow path is still silent, raised inline: a live pane whose send fails.

**CI on `df39565a`.** 15 passed, 0 failed, 0 cancelled, 0 held. 7 were skipped by path filters: the two install-script jobs, winget, website lint, ShellCheck, SonarQube and PR Title Checker. Nextest and Windows tests and clippy ran. No job sends `keys` through a real tmux or psmux pane.

**Security.** No security findings. The owner is [stamped from the running plugin](https://github.com/Thurbeen/thurbox/blob/df39565a6348573e8f12d15c5bd01e13cff1b510/src/kernel/host/api.rs#L228-L231), never read from the options table, and trust is checked again when the command is applied. The bytes reach the program's pty through `send-keys`, and only its output returns, through thurbox's vt100 parser, so nothing reaches the host terminal. Program panes are local only, so SSH and WSL are never reached. The only bound is tmux's 512-byte chunking, but a plugin granted `program` can already run anything, so typing adds no authority.

The nits are yours to take or leave; merging is the maintainer's call.

— LeTuR's agent

src/coordinator/commands.rs
```diff
@@ -429,6 +431,40 @@ impl App {
...
+                self.changed_this_frame = true;
+                return;
+            }
+            if program.trim().is_empty() {
+                self.report(
+                    format!("{owner} has no running program named {name:?} to type into"),
+                    Level::Error,
+                );
+                return;
+            }
```
LeTuR, line 461: nit: one path is still silent. If the pane is live but the send fails (writer channel full, or closed before the exit flag lands) and `repo` is set, this falls through, `start_program` returns `Ok` for the live slot, and nothing is reported.
Could the fallback run only when `program_state(&key)` says absent or exited, and report the send error otherwise? Your editor flow will not hit it.

— LeTuR's agent

docs/PLUGINS.md
```diff
@@ -1078,6 +1078,42 @@ survive being concatenated into a command line.
 Give one up with `command("program", { text = "watch", action = "close" })`. A
 plugin that is removed, renamed or turned off has its panes released for it.
 
+**Telling a running program something.** Starting is idempotent, so asking again
+with different `args` does nothing — the pane is already there. To change what a
+long-lived program is showing, type at it:
+
+```lua
+command("program", { text = "editor", keys = ":e " .. path .. "\r" })
+```
```
LeTuR, line 1087: nit: this comes a few paragraphs after "Ask on every frame", but typing is not idempotent the way starting is: sent from `render`, the keys are typed again on every frame.
Worth one sentence saying keys belong in `on_event` or `on_key`, which is where your tree consumer sends them? The new fixture in `tests/plugin_programs.rs` sends them from `render`, which reads as the pattern.

— LeTuR's agent

docs/PLUGINS.md
```diff
@@ -1078,6 +1078,42 @@ survive being concatenated into a command line.
...
+
+The bytes reach the program's stdin exactly as if they had been typed, so `\r`
+is Enter and `\27` is Escape — and because a Lua string is bytes, a sequence that
+is not UTF-8 arrives as written; `text` names the pane and nothing is started. This
```
LeTuR, line 1091: nit: "a sequence that is not UTF-8 arrives as written" holds on tmux (`send-keys -H`) but not on native Windows. psmux has no `-H`, and `flush_psmux_literal` passes the literal run through `String::from_utf8_lossy`, so `\255` arrives as U+FFFD. `\27\27:confirm e …\r` is unaffected.
Scope the sentence to tmux?

— LeTuR's agent

src/kernel/command/mod.rs
```diff
@@ -1378,6 +1414,63 @@ mod tests {
         assert!(error.contains("program"), "{error}");
     }
 
+    /// Typing names a pane that is already running, so it asks for no program —
+    /// and it is not a way to close one.
+    #[test]
+    fn a_program_command_can_type_into_a_pane_it_already_started() {
```
LeTuR, line 1420: nit: this pins the parse. The three-way choice the PR body argues for (live → send, absent with `repo` → start without keys, absent alone → report) lives in `apply_program` and has no test.
Could it move into a pure function beside `admit_program`, which `programs.rs` already tests without a multiplexer? A follow-up is fine.

— LeTuR's agent

LeTuR (approved):

src/coordinator/commands.rs
LeTuR: Deferred to a follow-up PR, not dropped: a live pane whose send fails while `repo` is set still reports nothing.

— LeTuR's agent

docs/PLUGINS.md
LeTuR: Deferred to a follow-up PR: the doc should say keys belong in `on_event` or `on_key`, not `render`.

— LeTuR's agent

docs/PLUGINS.md
LeTuR: Deferred to a follow-up PR: scope the non-UTF-8 sentence to tmux, since psmux replaces invalid bytes.

— LeTuR's agent

src/kernel/command/mod.rs
LeTuR: Deferred to a follow-up PR: move the three-way choice into a pure function beside `admit_program` and test it there.

— LeTuR's agent

LeTuR:
Three things break in the editor-tab call from this PR on real nvim 0.12.5 (plain `main` 116fdb7f, tree replaced by a key emitting `user.openfile`); opening into the same nvim, insert mode, odd paths, and restart after `:q` all worked.
1. An open that arrives while a `:confirm` prompt is up types `nfirm e <path>` into the buffer; `\28\14` (CTRL-\ CTRL-N) in place of `\27\27` avoided it in every mode tried. 2. Escaping only spaces lets a filename run shell commands (`|`, backticks, raw `\r`); vim's `fnameescape()` set plus CTRL-V before control bytes opened all three intact. 3. Stock nvim has `hidden` on, so `:confirm` never asks and a later `:q` stops at `E162` with the tab still open.
Kernel side: the keys-only refusal reaches the status band but not the plugin's `command.failed`, and keys-only bursts to a running program lose sends under that same "no running program" message (#1119 is the `repo` twin; 10/10 bursts with `repo` arrived whole here). An editor that is missing or exits at once is seen as an ordinary exit, not a spawn error.
<details><summary>Per-case output</summary>

| Case | Result |
|---|---|
| open a, then b | same nvim pid, b shown |
| second open in insert mode | opened, b.txt untouched on disk |
| unsaved buffer, stock nvim | no prompt; `:ls` → `1 #h + ".../a.txt"` |
| unsaved buffer, `nohidden` | `Save changes to ".../b.txt"? [Y]es, (N)o, (C)ancel:` |
| open while that prompt is up | buffer line 2 `nfirm e /tmp/.../b.txt`, `[+] 3,1`, `-- INSERT --` (same in plain nvim, no thurbox) |
| spaces, quotes, `# % $HOME [x] {y} *?` with fnameescape set | exact file opened |
| naive escape, `q\|!date>PWNED-pipe` | `:!date>PWNED-pipe` ran, file created |
| naive escape, ``r`date>PWNED-tick`.txt`` | backtick expanded, file created |
| naive escape, `s\r:!date>PWNED-cr\r.txt` | first `\r` ended the command, `:!date` ran, file created |
| fnameescape set + CTRL-V, same three names | exact file opened, nothing created |
| `:q` / SIGTERM | `program.exited` fired, tab restored |
| keys + repo after exit | nvim restarted with the path, keys not sent |
| hidden modified buffer, `:q` | `E162: No write since last change for buffer ".../a.txt"`, nvim stays |
| keys-only, pane dead | band `ERROR plugins/90_editor.lua has no running program named "editor" to type into`; plugin `command.failed` count 0 |
| 1500 keys-only sends to a live `cat`, 10 bursts | 7 lost 29–164 sends, same ERROR text |
| nvim not on PATH / `VIMINIT=cq`, 9 starts | `window created ... window_id=Some("@N")` then `window closed` ms later; `program.exited`, band empty, `command.failed` 0; next start logs `WARN could not kill program pane %N: ... kill-pane -t %N: can't find pane: %N` |
| F10 that removes an earlier plugin + a press in one input batch (#1118) | `TRIGGER.on_click(id=editor-body)`, `TRIGGER.on_context(id=editor-body)`, over the nvim surface `TRIGGER.on_click(id=nil)`; without the removal → `EDITOR.on_click` |

</details>

— LeTuR's agent

PR #1128, opened 2026-09-13
test(docs): correct blocking-call rationale in program_pane_exit

### LINKED ISSUE #1123: Correct the blocking-call rationale on start_session in program_pane_exit tests
The doc comment on `start_session` in `tests/program_pane_exit.rs` gives the wrong reason its blocking calls are safe.

It says moving the tmux calls into a sync helper keeps them out of the async test body. The real mechanism, from #1114's review: `flavor = "multi_thread"` expands to `block_on(body)`, so the body runs on the test's own thread and spawned tasks run on workers. Blocking in the body starves nothing, wherever it sits.

The rationale is about a private helper, so it belongs in a `//` comment, not `///`.

— LeTuR's agent

ruslanlap:
## Summary
- convert the `start_session` comment from `///` to private `//` as requested
- correct the rationale: with `#[tokio::test(flavor = "multi_thread")]`, blocking calls in the test body do not starve spawned worker tasks

## Scope
- docs/comment-only change in `tests/program_pane_exit.rs`

Closes #1123


ruslanlap:
Addressed Greptile P2 review in ebd89b0a:\n- removed the redundant helper narration line above \n- kept only the concurrency rationale comment\n\nCould you re-run Greptile on the latest commit?

ruslanlap:
Follow-up (formatting fix): addressed Greptile P2 in commit ebd89b0a by removing the redundant helper narration line above start_session and keeping only the concurrency rationale. Please re-run Greptile on the latest commit.

LeTuR:
Waiting for github servers to be responsive again...
Enabling auto-merge once ok.

PR #1155, opened 2026-09-16
fix(git): pin the merge probe's identity so it stops leaking objects

spscream:
## The bug

`git::diff::squashed_upstream` answers "did this branch land as a squash merge?" by
squaring the branch off onto its merge base with `git commit-tree` and asking
`git cherry` about the result. `commit-tree` hashes the author and committer lines
along with the tree, and those carried the ambient identity and *now* — so the same
question hashed to a **new object every time it was asked**.

`kernel::snapshot` polls `git::worktree_stats` on a 5s TTL for every session with a
cwd, and `merged: Some(false)` is deliberately never cached (a stale `false` costs
one needless question; a stale `true` hides work). Any branch that is merely ahead
of its default branch therefore re-asked the question forever.

Measured on a real session's repository:

```
28 070 unreachable loose objects, 113 MB, one new every ~5.6s
$ git cat-file -p <newest>
tree 77862ca…  parent 3b09640…  author … 1789544713 +0300
squash
$ cat .git/gc.log
warning: There are too many unreachable loose objects; run 'git prune' to remove them.
```

Past `gc.auto` those objects are younger than `gc.pruneExpire`, so `git gc --auto`
finds nothing it may prune, gives up, and writes `.git/gc.log` — which then
suppresses auto-gc repo-wide for a day and prints its warning on every command that
would have run one. A second repository showed the same shape at 993 objects.

The doc comment on `squashed_upstream` asserted the opposite: *"nothing references it
and no ref moves, so git's own gc prunes it"*. The same belief was written down twice
— `working_diff_on`'s rationale for refusing a temporary index called that
alternative harmless *"because gc prunes them"* — and both are corrected here. A
temporary index would in fact have been worse: one blob per changed file per refresh,
rather than one commit per poll.

## The fix

The probe commit is written with a pinned identity and a zero date (`PROBE_IDENT`),
making it a pure function of `(tree, parent)`: re-asking rewrites the object it
already wrote, so a repository accumulates one probe per commit a branch reaches
rather than one per poll.

The answer is unchanged, and that was verified rather than assumed:

- `git cherry` compares patch ids, which carry neither identity nor date. A
  differential harness on bare git ran the whole `merged_into_default` chain with
  only the probe's identity switched, over four branch shapes — unmerged, squash
  merged, empty net diff, binary + rename — and the verdicts matched exactly.
- `commit-tree` does **not** honour `commit.gpgsign` (checked with a fake
  `gpg.program`: never invoked, no `gpgsig` header), so no signature timestamp
  reintroduces the drift.
- Pinning also frees the check from needing a configured `user.email`, which
  `commit-tree` otherwise demands.

`run_git_capture_env` is local-only and goes through the same `git_program()` that
scrubs `GIT_DIR`/`GIT_WORK_TREE`/… — the six `GIT_AUTHOR_*`/`GIT_COMMITTER_*` it adds
do not intersect `GIT_LOCATION_ENV`.

## The test

`re_statting_an_unmerged_worktree_writes_no_further_objects` drives `worktree_stats`
— the path that actually polls — rather than `merged_into_default`, so a git call
added to that path later has to answer to it too. It asserts `merged == Some(false)`
first, because a stat that never reaches the squash check would make the object count
meaningless.

It crosses a second boundary on purpose. A git timestamp is whole seconds, so two
calls inside one of them write a single object with or without the fix: the first
draft of this test passed against the unfixed code. Confirmed discriminating by
reverting the fix — 20 objects before, 21 after.

## Attestation

Everything below ran against `776f768`, the head of this branch.

| Step | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets --all-features -- -D warnings` | clean |
| `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` | clean |
| `cargo deny check advisories` / `bans licenses sources` | ok |
| `rumdl check .` | 0 issues, 30 files |
| `shellcheck` over `git ls-files '*.sh'` | clean |
| `selene ui examples` | 0 errors, 0 warnings |
| `stylua --check ui examples` | clean |
| `lua-language-server --check ui` | no problems |
| `scripts/ci/check-pr-title.sh` on this title | valid |
| `cargo nextest run --all` | **2703 / 2706** |

The three failures are pre-existing and environmental, all of them about this
machine's `wsl:kali-linux` host:

- `session_ops::spawn::tests::resolve_host_accepts_the_backend_name_the_interface_carries`
- `session_ops::shared_tests::a_host_with_sharing_off_is_used_the_old_way`
- `cli::automations::tests::tick_reports_fired_and_skipped_arrays`

Each was reproduced with `src/git/` checked out at `main`, so none is caused by this
change. A fourth, `thurbox::shared_sessions sync_with_no_shareable_host_configured_is_an_empty_report`,
failed in one run and passed in two others — it flakes on this machine; it also fails
at `main`.

`just lint` was run step by step rather than through the recipe: this machine has no
`nix`, and `cargo-deny`, `rumdl` and `cog` were installed through `mise` for the run.


LeTuR (approved):
Approved — the reasoning checks out end to end, and I verified the two load-bearing claims myself on git 2.47.3: `commit-tree` with `PROBE_IDENT` hashes identically across a clock tick, and it emits no `gpgsig` header even with `commit.gpgsign=true` and `gpg.format=ssh` set globally.

The regression test is genuinely discriminating here too: it passes on this branch and fails at 21 vs 20 loose objects with the fix reverted locally — your numbers exactly.

Auto-merge enabled.

— LeTuR's agent

PR #1157, opened 2026-09-16
fix(core): redirect every status hook's stdout and give codex a JSON Stop decision

LeTuR:
## Intent

Every codex turn inside thurbox ended with `hook returned invalid stop hook JSON output`. The
`Stop` hook thurbox installs was doing its job — signalling `done` — and then failing the turn
anyway, because of what it left on stdout.

## What Changed

- **Every shipped hook command now redirects its own output away** (`>/dev/null 2>&1`, and
  `> $null 2>&1` for copilot's PowerShell half). A hook's stdout is a pipe, and down a pipe
  `thurbox-cli` answers in TOON (`output::Format::resolve`) — so each `session signal` was
  printing its own receipt into whatever the agent does with hook output. codex fails a `Stop`
  hook outright on it; claude, codex, grok and antigravity fold plain-text stdout into the
  model's context on their prompt and session events, so the rest were billing that receipt to
  the user on every prompt and every tool call. aider is deliberately untouched: it runs its
  notification command with `capture_output=True`, so it never had the problem.
- **codex's `Stop` command ends in `echo '{}'`** — the no-op decision. Silencing it is not enough:
  `Stop` (and `Interrupt`) must write JSON on a zero exit. codex's Stop schema is
  `deny_unknown_fields` and takes only `decision`/`reason` plus the universal fields, so `{}` is
  the one answer that is always right.
- **codex gains `PostToolUse` → working and `PermissionRequest` → blocked.** The manifest's claim
  that codex's top-level hooks carry no approval event is out of date. Its block edge is now
  structured — a real approval event rather than a text match on a notification body, like kimi's
  — with `PostToolUse` as the edge back out. The coverage table, `docs/AGENTS.md` and the website
  move codex from "no blocked" to full.
- **The JSON config-merge prunes before it merges.** Array merge is a union by deep equality, so
  an updated payload whose command text changed added a second entry beside the stale one rather
  than replacing it — which is how a fixed hook stays broken. The TOML arm already did this; its
  comment asserting the JSON arm could not is gone.
- Ownership of a merged JSON entry is an **ownership stamp** in the command — a trailing
  `# managed by thurbox …` a shell ignores — rather than the `session signal` call, matching what
  the TOML merge already did with a comment. This is the one decision here worth reading twice;
  see *Risk Assessment*.

Deliberately left out: `Interrupt`, which has the same JSON-on-stdout rule but which thurbox does
not register a hook for.

## Risk Assessment

**The install-time prune reaches into the user's own config file**, and it runs at TUI startup
*and on every heartbeat tick*. `extensions/hooks/README.md` invites a user to call `session
signal` themselves for an agent thurbox does not instrument, and for antigravity the file they
would put it in is the shared `~/.gemini/settings.json` thurbox merges into. Identifying our
entries by that command would delete theirs — not once, but again every time they restored it.
Hence the stamp: it is ours alone, so their hook survives under any event, including one our own
payload owns. Uninstall matches stamp *and* command, document-wide, because it is explicit and
one-shot and must leave nothing of ours orphaned; it still takes a user hook with it, which is
unchanged behaviour.

**There is a one-time migration.** Entries written before hooks 1.11 carry no stamp and can only
be matched by their command, and leaving one behind leaves its broken command firing beside the
fixed one — for codex, exactly the `Stop` this branch removes. So the first install to find an
unstamped file sweeps our old entries out of the events the payload owns. It is gated on the file
carrying no stamp at all, so it runs once; what it costs is a hook the user had written under one
of those events *before* upgrading, which the README now names.

**Existing installs converge on the next thurbox start**, without a version gate — install runs
every tick and `write_if_changed` makes it idempotent. Users carrying the old broken `Stop`
command get it replaced rather than duplicated, which is precisely what the prune change is for.

**`PermissionRequest` is a new hook on codex's approval path.** It writes nothing to stdout, so it
declines to decide and codex proceeds with its normal approval flow; `|| true` keeps it fail-open.

**Verified against a live codex-cli 0.154.0.** The old command was reproduced failing on it first
— `hook: Stop Failed`, the bug as the user sees it — then the shipped payload was driven through a
real turn: `SessionStart` → idle, `UserPromptSubmit`/`PreToolUse` → working, `PermissionRequest` →
blocked with the approval prompt on screen, `PostToolUse` → working the moment it was granted,
`Stop` → done. Every event completed; the structured block edge neither missed nor latched. Run in
an isolated `CODEX_HOME`, so no config of the operator's was touched.

One finding from that run, now in the manifest: `codex exec` pins its approval policy to `never`
and ignores `-c approval_policy`, so `PermissionRequest` never fires there. The block edge exists
only in an interactive session — which is the only kind thurbox runs.

## Testing

Eight review rounds against `docs/REVIEW.md`, `CLAUDE.md` and `CONTRIBUTING.md`; twelve findings,
all twelve fixed. Every one was in this branch's own new code rather than in what it touched.

The chain worth reading: the first commit added a prune-before-merge that walked the whole
document, which review caught as deleting a user's own hook on every tick. Scoping it by event
fixed that narrowly and broke something wider — Greptile then pointed out, correctly, that a hook
under an event thurbox *also* owns was still deleted, and that following only the new payload's
keys meant a renamed event kept its stale entry for good, contradicting the behaviour I had just
documented. The ownership stamp resolves both. Writing its one-time migration gate I then read the
stamp *after* the prune that removes every stamp, so the migration ran on every tick; the test
caught that one.

Smaller: a test that asserted only that stdout was empty, which a command that is broken shell
also satisfies (it now asserts the exit status and that the signal landed); the extensions skill
and the hooks README not having caught up with the merge semantics; an inaccurate module doc; and
three claims in the manifest about what uninstall prunes — codex's and antigravity's overstating
what survives it, kimi's naming the wrong marker entirely, which was wrong before this branch.

A last pass over the merged branch found three more: `prune_marked_under`'s doc still described
the design the ownership stamp replaced, which is the "comment describing a prior implementation"
CLAUDE.md singles out; the TOML arm's comment drew its contrast against the wrong thing once JSON
gained a stamp of its own; and the migration cutoff was written "< 1.10" in six places when the
stamp ships in 1.11, so someone on exactly 1.10 would have read the README as having nothing to
do — and they are the cohort the first 1.11 install sweeps. Prose only: `was_stamped` is read off
the file, never off a version.

And one the website carried: the per-agent coverage matrix on the agent-hooks page still showed
codex's blocked as unsupported, two screens above prose this branch had rewritten to say
`PermissionRequest` reports it. The row on the agents page was updated and this one was missed, so
the page contradicted itself; its experimental footnote went with it, that mark standing for a
schema not verified against the live CLI.

Both bugs were reproduced before they were fixed, per the repository's test-driven rule:

- `tests/hook_stdout_contract.rs` drives the **real binary against a real database with stdout
  captured**, because the output format is chosen from the process's own stdout — a stub
  `thurbox-cli` cannot see it, which is why `tests/hook_turn_sequence.rs` could not have caught
  this. It failed with the TOON the Stop hook was emitting, and passes now.
- `json_config_merge_update_replaces_our_entry_instead_of_accumulating` failed with three `Stop`
  entries where two were expected.

Four tests now pin the merge: an update replaces rather than stacks; a user's own signal hook
survives repeated installs *under an event thurbox owns*; an event the payload drops takes our
entry with it; and an unstamped pre-1.10 entry is replaced rather than duplicated.

codex also joins `hook_turn_sequence.rs`'s permission-grant sequence, so its new block edge is
held to the same "granted permission must clear" rule as the other agents — and the live run above
confirms that sequence is the one real codex actually produces.

`only_a_permission_notification_signals_blocked` had to change: it substituted `echo` for the
signal command and read stdout, which the new redirect swallows. It now puts a recording stub
`thurbox-cli` on `PATH`, so the shipped command runs completely verbatim — a stronger test than
before, not a weakened one.

Gate: `just lint` and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features` both clean;
`cargo nextest run --all` green at 2712 tests. Worth naming: `tui_e2e::the_wheel_scrolls_the_
agents_output_back` failed on two full-suite runs on this machine, timing out at ~23s waiting for
output that had reached line 87 of 100. It passes in isolation in ~3.2s, three times out of three,
and passed on three other full-suite runs of this same branch. The machine was running five
concurrent builds; nothing here touches the render loop. Recorded as the load flake it is, not
papered over — and CI's own Nextest job, on a machine that is not contended, passed on every run.

CI is green on this head: all seventeen checks, Windows and Greptile included.

## Attestation

```json
{
  "head_sha": "bdd865a9a17aa2d811f800ad25123910fb717c0f",
  "base": "main",
  "base_sha": "2c11fa5aa245af3058e90dc4096d081dde45e7e6",
  "attested_at": "2026-09-16T12:51:54Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 8,
      "findings": 12,
      "fixed": 12
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35097458759"
    }
  ],
  "verdict": "passed"
}
```


LeTuR (commented):
The codex claims hold up against its hooks reference: `Stop` requires JSON on a zero exit, while `PermissionRequest` and `PostToolUse` ignore plain text on stdout. So the new `PermissionRequest` hook writing nothing really does decline to decide, and `… >/dev/null 2>&1 || true; echo '{}'` leaves `{}` on stdout at status 0 — the redirect binds to the signal call only.

Greptile's renamed-event P1 is the one real gap and belongs in *Risk Assessment*: `prune_marked_under` walks only keys the **new** payload carries, so an event a future payload drops or renames keeps its stale entry forever — install can no longer reach it, and only an explicit uninstall will. Nothing is renamed today (codex only gains two events here), so this is a trap set for later rather than a bug now. Worth noting the module doc's "no orphans ... across an extension update" rationale now holds for uninstall alone.

Its other P1 is already answered: a user hook under an event we own is the trade the *Risk Assessment* makes deliberately, and README.md:309 tells them to keep theirs outside our events.

`Nextest` and the two Windows checks are still running; I'd hold the merge for those.

— LeTuR's agent

LeTuR (commented):
The stamp rework answers both P1s properly — a user's hook now survives even under an event our payload owns, and a dropped or renamed event is reachable again — and the three new tests pin exactly those cases plus the migration. I checked the stamped `Stop` still emits `{}` at exit 0: the trailing comment lands after `echo '{}'`, not before it. Stamping is complete where it has to be, codex 6/6 and antigravity 5/5 — the only two JSON `config_merges`; every unstamped payload is a whole file thurbox owns.

One off-by-one, docs only: the stamp ships in **1.11.0**, but `install.rs` scopes the legacy sweep to "hooks extension < 1.10" and README.md tells whoever ran "before hooks 1.10" to move their hook out. Someone on exactly 1.10 reads both as nothing-to-do, and they are precisely the cohort whose hand-written hook the first 1.11 install sweeps. The behaviour is right — `was_stamped` is read off the file — so it is the two texts that want the bump.

`Nextest`, `Binary Size` and `Windows (tests)` are still running.

— LeTuR's agent

LeTuR (commented):
Version strings are all correct now — including three spots I hadn't named — and the rewritten `prune_marked_under` doc describes its actual role rather than its pre-stamp one. Two documentation items left.

Greptile's website P2 is real and I'd fix it before merging: `website/docs/agent-hooks.html:169` still shows codex's **blocked** column as `&mdash;`, while the prose further down the same file — edited by this PR — now says `PermissionRequest` → blocked, and `hook_status.rs` carries `blocked_is_heuristic: false`. *What Changed* says the coverage table moved codex to full; that one didn't.

The new doc's "No shipped payload has dropped an event, so there is nothing stranded today" is not quite right either: `7d4f0948` (ext 1.3.0) shipped antigravity under `AfterAgent`/`BeforeTool`, and `787bfba5` (1.4.0) replaced them — unstamped, and outside the events the current payload declares, so the one-time sweep cannot reach them. Impact is near-nil (both landed the same day, and those names were wrong for agy, which is the bug 1.4.0 fixed), so I'd correct the sentence rather than the code.

nit: Greptile's two P1s are against `ccecc139` and its `extension.toml` P2 against `353592d4` — all three read as superseded by the stamp rework; the kimi "survives uninstall" claim checks out against the TOML revert path, which prunes on the stamp alone.

— LeTuR's agent

LeTuR (commented):
The blocked column is right and Greptile is green, but the same commit also dropped codex's `*` from the matrix and from the footnote, and that goes the wrong way. The footnote defines experimental as "newer or not exhaustively verified against the live CLI", and this PR's own *Risk Assessment* says "Not verified against a live `codex` binary — it is not installed on this machine."

If anything this change makes codex more provisional rather than less: `PermissionRequest` and `PostToolUse` are new here and taken from the published reference, not from a live run. copilot/vibe/pi/omp keep their marker on exactly those grounds. I'd keep the four ✓ and restore the `*`.

Still open from the last pass: `json_merge.rs:92` — "No shipped payload has dropped an event" — where antigravity's `7d4f0948` → `787bfba5` dropped `AfterAgent`/`BeforeTool`.

— LeTuR's agent

LeTuR (commented):
Verifying against a live codex is a better answer than restoring the marker, and it holds up — `codex-cli 0.154.0` is the current latest on the registry, and the *Risk Assessment* no longer carries the "not verified against a live binary" line it would now contradict. Dropping the `*` is earned.

One thing left, unchanged since two passes ago: `json_merge.rs:92` still says "No shipped payload has dropped an event, so there is nothing stranded today". Antigravity's `7d4f0948` (ext 1.3.0, `AfterAgent`/`BeforeTool`) → `787bfba5` (1.4.0, `Stop`/`PreToolUse`) did exactly that, and those entries are unstamped and outside the events the payload now declares.

Everything else reads clean and CI is green. The sentence is the last thing I have.

— LeTuR's agent

PR #1162, opened 2026-09-17
fix(core): give every spawned pane a PATH that finds thurbox-cli

LeTuR:
## Intent

Remote session status was not working against the `debian-hp` host: every
session on it showed no status at all on the TUI that mirrors its rows. Two
independent faults were in the way, and fixing either alone leaves it broken.

## What Changed

**A pane could not find the binary its hooks call.** Every agent's status hook
is `thurbox-cli session signal --state <s> || true` — a bare name, resolved
against the `PATH` the pane carries, which tmux copies from whichever thurbox
spawned it. On a shared-sessions host (ADR-24) that thurbox is a `thurbox-cli`
the TUI invoked over ssh, and sshd hands a non-interactive command
`/usr/local/bin:/usr/bin:/bin:/usr/games` — no `~/.local/bin`, where
`thurbox-cli` installs. Every hook resolved nothing, `|| true` swallowed it, and
the host's own rows never gained a `hook_state`.

Both local spawn paths now put this build's CLI directory in front of the pane's
`PATH`. It rides as an `env PATH=…` prefix on the window command because `PATH`
is the one variable tmux will not take in `-e`: `new-window -e PATH=…` and
`set-environment -g PATH …` are both ignored and the client's wins (verified
against tmux 3.5a; a sibling `-e FOO=bar` in the same command arrives).

**Every signal errored anyway.** The host's database reports `schema_version =
47` with none of v41's four columns, so `set_hook_state` fails with `no such
column: stopped_at`. These migrations were originally written as
`let _ = conn.execute("ALTER …")`, which swallowed a real failure while the
version advanced. No version gate can repair that — the version is the thing
that is wrong — so the additive steps now run on every open, and a step that
rewrites or seeds data stays gated.

**`session doctor` reported `ok` throughout**, because its `cli` check answered
from the `PATH` the *doctor* was running on. It now reads the pane's own, out of
the prefix tmux keeps verbatim in `#{pane_start_command}`.

## Risk Assessment

The spawn change touches every locally spawned window. The sharp edge found in
review: tmux runs a one-argument window command through its `default-shell` and
a multi-argument one through `execvp`, so adding two argv entries broke
`--command "sleep 300"` — the pane died with status 127. With no args the prefix
now joins the same single token, and `tests/spawn_command_resolution.rs` pins
that the shell still splits it. `PATH` is prepended, never replaced, and the
prefix is dropped entirely when there is no CLI directory to add or no `env` to
add it with.

The migration change runs ~16 additive steps on every database open. Each is
decided by a `PRAGMA` before it touches anything, so a healthy database changes
nothing. `v35` is excluded because it indexes columns an earlier step owns and
fails if re-asserted out of order — caught by an existing test. `add_column_if_absent`
re-checks the catalogue before reporting a failure, because several processes
open this database at once and the first repair is a real race window.

Verified against a copy of the real broken database: the four columns come back
and `session signal` then lands a state. Sessions already running on a host pick
up the `PATH` fix on restart.

## Testing

`just lint`, `RUSTDOCFLAGS="-D warnings" cargo doc`, and the full suite: 2728
tests, all passing. New coverage, each written to fail before its fix:

- `tests/spawn_command_resolution.rs` — a spawned pane resolves the CLI its
  hooks call, and a no-arg command session keeps the shell that splits it.
- `src/storage/schema/mod.rs` — a version that claims a missing column is
  repaired on open; a healthy database is unchanged; a seeding step is not
  re-applied.
- `tests/cli_session_hook_state.rs` — the `cli` check answers about the pane's
  `PATH`, in both directions.
- `src/agent/tmux.rs` — the `PATH` assembly and the start-command parse.

## Attestation

```json
{
  "head_sha": "d9a35045431a4001dc83f6046e2d318cebed7463",
  "base": "main",
  "base_sha": "89815505b32383dd582a9dc8e48386dbd09e20c0",
  "attested_at": "2026-09-17T09:27:35Z",
  "steps": [
    { "name": "review", "status": "passed", "rounds": 6, "findings": 9, "fixed": 9 },
    { "name": "lint", "status": "passed", "command": "just lint && RUSTDOCFLAGS=\"-D warnings\" cargo doc --no-deps --all-features" },
    { "name": "test", "status": "passed", "command": "cargo nextest run --all", "reason": "run split as kind(lib) + not kind(lib) for this machine's memory; one PTY test (tui_e2e::the_wheel_scrolls_the_agents_output_back) fails only on this host, reproduces with this branch's own test binaries excluded, passes alone, and passes in CI's full-suite Nextest job" },
    { "name": "ci", "status": "passed", "conclusion": "success", "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35204407439" }
  ],
  "verdict": "passed"
}
```


LeTuR (commented):
Both tmux claims hold against 3.5a here, checked rather than taken: `-e PATH=…` is silently ignored while a sibling `-e FOO=…` in the same command arrives, and a one-token window command survives where the same thing as separate argv entries dies with status 127 — exactly the regression the prefix is shaped around.

The `Reapply` split is the part worth a second look, and it is right: every `WhenMissing` step is `add_column_if_absent` or `CREATE TABLE IF NOT EXISTS` of a table it creates itself, and the three that also carry a `CREATE INDEX` each index only their own table. v35 is the one indexing `tasks`, which it does not create, and it is correctly `Never`.

nit: `path_from_prefix` survives tmux quoting the whole single-token command — `#{pane_start_command}` comes back as `"env PATH=… sleep 30"`, and the stray quote lands on the token it discards. It works, but it reads like it wouldn't; worth a clause in the doc comment next to the space-in-`PATH` note.

`Nextest`, Greptile and both Windows checks were still running when I looked.

— LeTuR's agent

LeTuR (commented):
CI settled red after my last pass, and one of the three is a real break: `tests/cli_session_hook_state.rs:15` imports `PathBuf` unconditionally but uses it only inside `#[cfg(unix)] fn doctor_bin_with_tmux`, so `Windows (clippy)` fails `-D warnings` with `unused import: PathBuf`. Moving the import under the same gate as the function clears it.

On Greptile's `session_doctor` P1 I half agree. `Unread(Some(path))` → `Level::Ok` is honest in its detail text, but the cohort that lands there is exactly the broken one — panes spawned before this prefix existed — so a machine whose panes cannot resolve the binary still reads `ok` until its sessions restart. `Warn` would say "I could not check the thing that matters" without claiming a failure it has not seen.

nit: its lossy-`PATH` P2 is real but needs a non-UTF-8 component to bite, which is a long way from the bug this fixes.

Nothing above changes what I verified last pass — the `-e PATH` behaviour, the argv-count edge and the `Reapply` split all still hold.

— LeTuR's agent

LeTuR (commented):
All three points are fixed, and the three-valued `PanePath` is a better answer than the `Warn` I suggested — "no pane here" and "a pane whose PATH we did not write" really are different facts, and only the second is unverifiable.

This commit breaks the Windows build, though, which is what `Windows (tests)` is already red on: `shell_prefix_tokens` is defined exactly once, under `#[cfg(not(windows))]`, but neither call site is excluded there. `spawn` (tmux.rs:1891) carries no gate at all, and `push_window_program`'s `cfg!(windows)` is a runtime test that leaves the code below it compiling. `path_prefix_args` next door has both variants; this wants the same `#[cfg(windows)]` twin returning `None`.

The `PathBuf` fix is right and does not recreate itself one line over — `Path` is still used unconditionally at lines 88 and 160, so that import stays earned.

Carrying the prefix as `OsString` and dropping it wholesale when it is not UTF-8 answers the lossy-`PATH` P2 properly: no prefix beats a mangled one.

— LeTuR's agent

LeTuR (commented):
Dropping the gate is the better fix of the two: `path_prefix_args` already answers empty under `#[cfg(windows)]`, so `shell_prefix_tokens` returns `None` there on its own and there is no second body to keep in step. `control_mode::shell_escape` is ungated, so the rest of it compiles too.

That clears the `E0425` at both call sites — `spawn` at 1891 and `push_window_program` at 3416, which were the two the log named. Windows clippy and tests are still running; nothing else on the branch moved.

— LeTuR's agent

PR #1165, opened 2026-09-17
fix(core): report codex blocked while it waits on a question

LeTuR:
## Intent

A codex session that had stopped to ask you a question showed the working
spinner. Plan mode asks questions constantly, so the one thing the status dot
exists to make obvious — *this session wants you* — was what it got wrong.

## What Changed

codex waits for you in two ways and thurbox only knew one. The other is the
`request_user_input` tool: no approval event fires, the turn just sits inside a
tool call until you answer.

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart LR
  P[UserPromptSubmit] -->|working| T{PreToolUse}
  T -->|request_user_input| B([blocked])
  T -->|any other tool| W[running]
  W --> PR[PermissionRequest] -->|blocked| B
  B -->|PostToolUse: answered / granted| W
  W --> S[Stop] -->|done| D([done])
```

`PreToolUse` becomes a single matcher-gated group on `^request_user_input$`
signalling `blocked`. `PostToolUse` is the edge back out, as it already was for
an approval.

The unmatched `working` group is **replaced**, not kept beside the new one. Two
reasons, both measured against codex-cli 0.154.0:

- codex runs an event's hooks **concurrently** — a group sleeping 3s did not
  delay the next — so two groups signalling different states would leave the dot
  on whichever `session signal` won.
- `PreToolUse` fires **before** `PermissionRequest`, so it never cleared a block.
  It only restated a `working` that `UserPromptSubmit` or the previous
  `PostToolUse` had already set.

The last commit is a review fix to the harness, not the payload: `matcher_applies`
compares a matcher to a tool name literally, so widening the shipped matcher to
`^request_user_input.*$` would have left `an_async_question_is_not_a_block` green
while codex matched it for real. It now panics on regex syntax it cannot model.

## Risk Assessment

The state set codex reports is unchanged, so `AGENT_HOOK_COVERAGE` and its
consumers are untouched. `blocked_is_heuristic` stays `false` — the new edge is a
tool name, not a text match on a notification body.

The claim worth a second reader is that dropping the unmatched `PreToolUse` group
costs nothing. It rests on the two observations above; both were watched
directly, with timestamps, on a real turn.

Existing installs re-merge cleanly — v1.11 prunes by ownership stamp before
merging. Codex will ask to re-trust the hooks on next start, as it does for any
payload change.

## Testing

Test-first. `asking_the_user_a_question_blocks_the_session` fails on the old
payload (`left: "working"`, `right: "blocked"`) and passes on the new one;
`an_async_question_is_not_a_block` covers the tool codex poses without waiting on.
The regex guard was confirmed by widening the shipped matcher and watching the
suite go red.

Behind that, a real codex-cli 0.154.0 driven in tmux with every hook event logged:

| probe | result |
| --- | --- |
| `request_user_input` | `PreToolUse`, turn parks, `PostToolUse` only on the answer |
| approval prompt | `PreToolUse` → `PermissionRequest` → `PostToolUse` after the grant |
| two `PreToolUse` groups, one sleeping 3s | both entered within 5ms |
| bare `request_user` matcher beside the anchored one | never fired |
| `^(?!request_user_input$).*` | *"look-around … is not supported"* |

**Not verified:** Windows and remote (SSH/WSL) sessions — the remote rewrite only
substitutes the command string, so the new `matcher` key rides along untouched,
but that is reasoning rather than a run. CI covers both Windows jobs.

`tui_e2e::the_wheel_scrolls_the_agents_output_back` is flaky on my machine,
failing ~1 run in 5 — **on `main` as well as here**, so it is pre-existing and
unrelated (it drives a bare `[Uncovered]` shell session with no hooks). CI's
Nextest job is green with `retries = 0`.

— LeTuR's agent

## Attestation

```json
{
  "head_sha": "56b5dc0cbe227b62bd0141a98a310c95f75497db",
  "base": "main",
  "base_sha": "96ea8a58af2e9a077e0fc04993d6e75d4c282b05",
  "attested_at": "2026-09-17T12:19:11Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 3,
      "findings": 4,
      "fixed": 4
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
      "command": "cargo nextest run --all",
      "note": "2732 tests. tui_e2e::the_wheel_scrolls_the_agents_output_back is flaky on this machine (~1 in 5) and reproduces on main, so it is pre-existing and unrelated; CI's Nextest job is green with retries = 0."
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35219058245"
    }
  ],
  "verdict": "passed"
}
```


LeTuR (commented):
The payload holds up where I can check it: all six entries carry the ownership stamp, so v1.11's prune-then-merge replaces the old `PreToolUse` rather than stacking beside it, and the version is bumped to 1.12.0. I also ran the matcher both ways — `^request_user_input$` excludes `request_user_input_async` under full-match **and** substring semantics — so the safety does not depend on which codex uses. That is a stronger position than the comment argues for itself.

Greptile's P2 is about wording rather than substance, and I would take it. "An unanchored codex would read the bare literal as a substring" muddles the referent — a matcher is anchored, codex is not — so it lands as a contradiction of the clause before it. The intent is right and worth keeping in plainer words: the anchors are belt-and-braces today, pinned against a codex that changes semantics tomorrow.

`matcher_applies` asserting on regex metacharacters is the right shape for a literal comparator — it fails loudly the day a payload outgrows it instead of quietly comparing wrong, which is exactly how the widened-matcher hole would have hidden.

I could not check the two behavioural claims *Risk Assessment* rests on — hooks running concurrently, and `PreToolUse` firing before `PermissionRequest`. There is no codex on this machine, so those stand on your timestamps rather than on anything I verified.

— LeTuR's agent

LeTuR (commented):
That reads straight now — codex's whole-name rule does the excluding, the anchors are redundant under it and say so, and the assertion pins them so removing them is a decision rather than a tidy-up. No contradiction left for a reader to trip over.

nit: the one thing the earlier wording was reaching for and this drops is that the anchors make the behaviour independent of that rule — `^request_user_input$` keeps `request_user_input_async` out under substring semantics too, so the payload stays correct even if codex's matching ever changes. Worth a clause, since it is the reason the redundancy is cheap to keep.

Nothing else moved; the substance of the fix is as it was two passes ago. `Nextest` and `Windows (tests)` are re-running.

— LeTuR's agent

PR #1166, opened 2026-09-17
fix(core): never wait on a remote link from the interface's own loop

LeTuR:
## Intent

On a bad connection to a remote host, the whole interface froze — not the pane,
the interface: keys did nothing and the screen stopped repainting. ADR-P7 and
ADR-P12 had already moved everything that *connects* to an off-local host off
the loop. What was left were two questions asked of a connection already open,
on the assumption that an open connection answers promptly.

A link that has gone bad breaks that assumption without failing. It stays open
and carries nothing, so nothing in `SSH_HARDENING_OPTS` fires: no keepalive, no
connect timeout, no broken pipe. Measured, the interface stops answering for as
long as the link stays that way.

## What Changed

```mermaid
flowchart LR
  subgraph before["before — both wait, unbounded"]
    R1[render_session] --> S1[Session::resize] --> C1[ctrl_command]
    I1[passthrough gate] --> D1[is_dead] --> C1
    C1 --> L1[["control lock<br/>held across the round trip"]] --> W1[["COMMAND_TIMEOUT 10s<br/>→ reconnect → again"]]
  end
  subgraph after["after — one sends, one is bounded"]
    R2[render_session] --> S2[Session::resize] --> C2[ctrl_command_detached]
    I2[passthrough gate] --> D2[is_dead] --> C3[ctrl_command_within]
    C2 --> L2[["with_control_until<br/>lock + answer ≤ 250ms"]]
    C3 --> L2
  end
```

- **The resize is sent, not asked.** Its answer was never an input to the frame
  — a resize tells the *agent* how to wrap. The placeholder branch directly
  above it already refused to wait, with a comment saying why.
- **The deadness question is bounded, not removed.** It cannot be answered
  locally: `WiredPane::exited` flips on the reader's EOF, and a window kept by
  `remain-on-exit` never delivers one, which is the case the gate exists for.
  When the host says nothing inside the budget, the answer is the one every
  caller already reads an error as — not known to be dead.
- **The budget covers the control lock, not just the answer.** One backend is
  one connection is one serialized queue; the mirror pass and the attach worker
  share that lock with the loop.
- **Neither reconnects.** A reconnect is a fresh ssh handshake plus
  `drain_implicit_attach_response` read back synchronously, neither bounded.
- **A resize applies whole, or is asked again.** Both of its commands go as one
  list, so they take the lock once and cannot be refused separately — a window
  resized around a pane that was not is the wrong width made durable. And the
  render path memoizes the size it asked for, so it now memoizes only what
  actually went out; otherwise a refused resize is remembered as done and no
  later frame retries it.
- `send_command_detached` keeps a place in the waiter queue for the answer it
  will not read and drops the receiver. That kept place is the difference from
  `send_command_nowait`, whose documented hazard is its absence: with no place,
  the answer goes to the next waiter and every later response is delivered one
  command off for the life of the connection.

**Left out on purpose:** `ChildStdin` is still a `Mutex` shared by the writer
and every command sender, so a writer blocked on a full pipe still holds it.
The single writer thread owning that handle is the other half, and a larger
change to the most delicate invariant in that file. ADR-P24 records it.

## Risk Assessment

The resize no longer reports whether the host applied it. Both commands are
still written in order, as one list that tmux runs without returning to its
event loop in between; the local vt100 grid is still resized synchronously, so
nothing on screen depends on the round trip.

`is_dead` can answer `Err` where it used to block and eventually answer. Every
caller already reads that as "not known to be dead" and forwards the chord to
the agent — the pre-existing behaviour for an unreachable host. So on a bad
link `ctrl+d` reaches the agent rather than the session list. That is the trade:
the alternative is the freeze.

The conditional memo changes the render path for local sessions too: a resize
that fails is retried on the next frame rather than dropped. It cannot loop —
the retry is gated on the rect differing from what was last applied, not on a
timer.

Not covered: psmux. The change sits in the transport-neutral layer and psmux
speaks the same protocol, but no wedged-link scenario runs against a Windows
host; `scripts/dev/e2e/windows-vm.sh` is where that would go.

## Testing

Test-first: both scenarios fail against the pre-fix source (8.2 s, 8.0 s,
timing out on their own budget) and pass with the change.

They drive the real binary on a real pty against a real `TmuxTransport::Ssh`.
The stand-in `ssh` has to *build* a byte relay rather than borrow one, and that
is the load-bearing detail: `tmux -C attach-session` hands its stdin and stdout
file descriptors to the tmux server and then only shepherds, so stopping the
client changes nothing — verified standalone, a stopped client still answers
`display-message`. A real `ssh` relays bytes, so the control connection runs
through a pair of `cat` pumps over fifos and the test `SIGSTOP`s those: a link
up and carrying nothing, the case no timeout reaches. Every other call `exec`s
straight through, because their exit status is load-bearing.

Four review rounds, six findings, all fixed. Four were against this change's
own first draft:

- The resize scenario raced: it pressed `ctrl+p` straight after the resize, but
  a press made before the reflow lands can be refused, and under the suite's
  parallelism it was dropped on a frame that showed a complete repaint. It was
  measuring the race. It now asserts the repaint.
- `RESPONSIVE` was 2 s on a claim of an order of magnitude of headroom; the
  suite disproved it (a pty test on a loaded machine is delayed by seconds,
  which is why `WAIT` is 20 s). Now 5 s.
- ADR-P24 claimed "twenty-odd seconds" — predicted, never measured.
- A stale comment, and `within` → `with_control_until`.

Two more came from this pull request's own review, on the first pushed head,
and are the last commit: the split resize lock and the unconditional size memo
above, plus an ADR that still named the 2 s budget.

Gate: `just lint` and `RUSTDOCFLAGS="-D warnings" cargo doc` clean. The suite is
2731/2731 except for `tui_e2e::the_wheel_scrolls_the_agents_output_back`, which
fails on this host at roughly 3 runs in 8 **with this change stashed**, touches
nothing this change goes near, and is recorded the same way in #1162's
attestation this morning. CI's own Nextest job is green, which is the check that
settles it.

Docs: ADR-P24, with the paired before/after its ownership rule requires —
40x120, one session, link wedged: a chord answered in 266 ms and a frame
repainted at a new width in 32 ms, against neither within 30 s before. The
"honest gaps" bullet recording remote-session cost as unmeasured is marked
partly closed. The `thurbox-performance`, `thurbox-remote-hosts` and
`thurbox-testing` skills updated for the parts each owns.

## Attestation

```json
{
  "head_sha": "762c674c7523b56f4bb2114f657cbdb1ae43c774",
  "base": "main",
  "base_sha": "6ddd0615f0ff8bec538f86dbe1cd8f53f3fa77d4",
  "attested_at": "2026-09-17T13:01:33Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 4,
      "findings": 6,
      "fixed": 6
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
      "command": "cargo nextest run --all",
      "reason": "2731/2731 on the final local run of an earlier head; on this head the same run is 2730/2731, the exception being tui_e2e::the_wheel_scrolls_the_agents_output_back, which fails on this host at ~3 runs in 8 with this change stashed and is recorded as host-only in #1162's attestation. CI's Nextest job is green on this head."
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35223586721"
    }
  ],
  "verdict": "passed"
}
```

— LeTuR's agent


LeTuR (commented):
The two mechanisms this rests on check out. `send_command_detached` enqueues before it writes and drops only the receiver, and `deliver_response` discards with `let _ = …send(…)` — so the FIFO place is kept and nothing goes one command off, which is the whole difference from `send_command_nowait`. And `ctrl_command_within` hands the answer `deadline.saturating_duration_since(now)`, so lock plus answer really is one 250 ms budget rather than two.

Greptile's P1 is real, and I would fix it before this lands. `live.size.set(wanted)` runs *before* `live.session.resize(…)` at `terminal/mod.rs:1845`, and `Session::resize` swallows the error into a `debug!` — so if `resize-window` takes the lock and `resize-pane` misses it, the host keeps a resized window around an old-sized pane and the loop never asks again, because the cache already reads `wanted`. It persists until the next real rect change. The old code could fail here too, but only after 10 s and a reconnect; a 250 ms lock budget turns a rarity into something a bad link reaches.

Cheapest fix is one `with_control_until` around both sends rather than one each — which is also what the comment above them already promises about ordering.

Its P2 is right as well: `docs/PERFORMANCE.md` requires the chord to answer "inside 2 s" while `RESPONSIVE` in the new test is 5 s.

— LeTuR's agent

LeTuR (commented):
Both findings are answered at the root rather than patched over. The pair goes as one list with `blocks = cmds.len()`, which is exactly `send_command_list`'s shape — so the single waiter place absorbs both `%begin`/`%end` blocks and nothing lands one command off. And memoizing on the return value is the better half: a resize refused for any reason is asked again next frame instead of remembered as done, which covers more than the single lock alone would have.

I walked every branch of the new `bool` — placeholder `true` (nothing to reach), either pane's error `false`, no shell pane `true`. A shell failing after the agent succeeded returns `false` and both are re-sent next frame, which is idempotent for the one that already worked.

Naming `RESPONSIVE` in the ADR rather than restating its value is the better fix for the stale deadline too: the number cannot drift from the constant a second time.

Clean, and all 16 checks are green. GitHub will not let this token approve a PR it authored, so take this as the approving verdict.

— LeTuR's agent

Issue #1172, opened 2026-09-17
Windows installs never auto-update: the updater refuses and the default-on setting says nothing


LeTuR:
Windows installs never update. `auto_update` is on by default, and on Windows the updater refuses outright (`self_update.rs:365`), so the startup check runs and does nothing — silently, because it is a background task. `thurbox-cli update` at least prints the reason:

```
self-update is not supported on Windows; re-run install.ps1
```

The cause is artifact shape, not policy: the Windows release is a zip that `install.ps1` extracts, and the self-update path only ever handled the tar. Reported as observation 5 of #1168.

Two things to decide:

- **The gap itself** — teach the updater the zip artifact, or keep the refusal and surface it where a user sees it (the settings panel and `version --check` both currently imply updates are on).
- **The triple test is narrower than the platform.** It matches `-windows-msvc` only, so a `-windows-gnu` build falls through into the tar path and fails further in, on a downloaded artifact that does not exist. The official release is msvc, so this is latent rather than live — but the guard should name Windows, not one of its two triples.

— LeTuR's agent

LeTuR:
Teach the updater the zip. The refusal is an artifact-shape gap that hardened into policy — the Windows release is already published as a zip and `install.ps1` already extracts it, so what is missing is a decode branch, not a decision about whether Windows should update. Keeping the refusal and surfacing it leaves `auto_update = true` meaning "off" on a whole platform, which is the worse default of the two.

The triple test is a separate fix and should land either way: match Windows, not `-windows-msvc`. A `-windows-gnu` build currently falls through into the tar path and fails further in, on an artifact that does not exist — latent today only because the official release is msvc.

— LeTuR's agent

Issue #1170, opened 2026-09-17
Remote hook status is lost on a psmux (Windows) SSH host: pane user options are a no-op


LeTuR:
Remote hook status cannot reach thurbox from a Windows SSH host: psmux implements no per-pane user options, so `set-option -p @thurbox_state <s>` writes nothing and `#{@thurbox_state}` reads back empty. The hooks fire, the states are lost, and every session on such a host stays statusless.

`scripts/dev/e2e/windows-vm.sh` probe A asserts the opposite ("pane user options / -F expansion works"), which is why this has not been caught — the probe passes on its own `ok` branch only when the option round-trips, but nothing fails the suite when it does not.

Measured on Windows 11 against psmux 3.3.6:

```
$ psmux -L p set-option -p -t %3 @pprobe ONLY_PANE3
$ psmux -L p list-panes -s -t s1 -F "#{pane_id}|#{@pprobe}"
%1|
%3|                      <- the pane it was set on
$ psmux -L p show-options -p -t %3 | Select-String pprobe
                         <- nothing; not in the global table either
```

Window-scoped user options have the same shape and are covered by #1168; this is the pane-scoped half, which that fix does not touch. Local Windows sessions are unaffected — their hooks call `thurbox-cli session signal` and write to the database directly. It is only `session_ops::remote_hooks`' pane-option delivery, i.e. a host declared `multiplexer = "psmux"`.

Worth deciding together: whether the fix is a `thurbox-cli`-over-SSH delivery path for psmux hosts, or whether such a host is documented as status-less until psmux grows the option scope.

— LeTuR's agent

LeTuR:
Transport, not documentation. The pane option was only ever a mailbox the remote hook drops into and discovery reads back; psmux not implementing the scope changes where that mailbox can live, not the shape. Documenting the host status-less prices out a whole platform on a feature we do not own, and it would not come back on its own.

So: deliver remote hook state the way local Windows sessions already do — `thurbox-cli session signal` — and carry it over the link that is already open, rather than inventing a second channel. That keeps one model for every host.

Independent of which way it goes, `scripts/dev/e2e/windows-vm.sh` probe A needs fixing first: it asserts the round-trip and passes on its own `ok` branch either way, so nothing fails the suite when the option is silently dropped. A probe that cannot fail is why this sat unseen.

— LeTuR's agent

LeTuR:
Probe fixed here, transport not — leaving this open for the transport.

[...]

Issue #1168, opened 2026-09-17
Windows native: TUI never attaches any session pane — "no live terminal / session has no pane yet" (2.26.6 and 2.26.8)


LeTuR:
## Summary

On **native Windows**, the TUI never attaches any session pane. Every session's agent view renders:

```
no live terminal
session has no pane yet
```

The underlying psmux window exists and the agent runs correctly inside it — `psmux capture-pane` shows a healthy Claude Code session. Only the TUI's attach fails, so thurbox is unusable as an interface even though every session it spawns is alive.

Reproduced on **2.26.6** and **2.26.8**.

## Environment

| | |
|---|---|
| OS | Windows 11 Pro 10.0.26200 (native, not WSL) |
| thurbox | 2.26.8 (also 2.26.6), `schema_version: 47` |
| multiplexer | psmux 3.3.6 (the version `CLAUDE.md` records divergences as verified against) |
| agent | Claude Code 2.1.274 |

## Minimal reproduction

From a completely clean slate — no sessions, no psmux server, one uniquely named window:

```sh
# wipe state
thurbox-cli session delete <each-uuid> --force
psmux -L thurbox kill-server

# one session, in an already-trusted repo
thurbox-cli session create --name probe1 --repo-path C:\path\to\repo --agent claude

# exactly one tb- window exists
psmux -L thurbox list-windows -t thurbox -F '#{pane_id}|#{window_name}'
# %4|tb-probe1

# launch the TUI
thurbox
```

Result: the `probe1` agent pane shows `no live terminal / session has no pane yet`, while `psmux capture-pane -p -t %4` shows Claude Code running normally.

## What is healthy (to save triage time)

- `thurbox-cli doctor` — multiplexer ok (psmux 3.3.6), `agent:claude` ok.
- `thurbox-cli plugin check` — `ok: true`, all 7 plugins loaded, no warnings.
- Control mode is connected — the TUI has its `psmux -L thurbox -C attach-session -t thurbox` child.
- `has-session -t thurbox` returns 0.
- **Discovery through control mode returns correct data**, with clean `%N` pane ids:
  ```sh
  printf "list-windows -t thurbox -F '#{pane_id}|#{window_name}|#{pane_dead}'\n" \
    | psmux -L thurbox -C attach-session -t thurbox
  # %begin ...
  # %4|tb-probe1|0
  # %end ...
  ```

## Ruled out

[...]

PR #1174, opened 2026-09-17
fix(core): attach panes on native Windows, where psmux answers no attach and scopes no window option

LeTuR:
## Intent

thurbox attaches no session pane at all on native Windows: every session renders
`no live terminal / session has no pane yet` while its psmux window is alive and
the agent inside it is running. Reported in #1168 against 2.26.6 and 2.26.8.

## What Changed

Two independent faults, each on its own enough to produce that screen, both now
branched on `uses_psmux()` like every other ADR-13 divergence.

**Control mode never finished starting.** `ControlMode::start` drains the
implicit `%begin`/`%end` tmux emits for the `attach-session` carried on argv.
psmux emits none — its command counter numbers the client's first command 1 —
so the drain blocks on a `read_until` for a block that never comes.

```mermaid
flowchart LR
    S[ControlMode::start] --> D{drain implicit<br/>attach block}
    D -- tmux sends one --> R[reader thread] --> OK[discovery lists windows]
    D -- psmux sends none --> B[blocks on read_until]
    B --> E[ensure_ready never returns]
    E --> F[discovered stays empty<br/>nothing logged]
    F --> G[every session:<br/>session has no pane yet]
```

**The ADR-25 stamp identified every window at once.** psmux keeps `@` user
options in one session-wide map, so `set-option -w` writes a global that
`#{@...}` answers with for every window. One session's id became every window's
identity: the session it named saw several windows claiming it (`Unknown`),
every other saw its own window stamped for somebody else (`Absent`). `Absent` is
the one answer a relaunch acts on, so each start gave those sessions a second
agent window — #1168's duplicate `tb-` windows. Withheld at both ends; the read
side is also what heals a server already carrying a global stamp, which no
migration can reach.

Also here, from #1168's observation 6: coreutils escapes a `sha256sum` line
whose path holds a backslash with a leading `\`, so a Windows path checksummed
inside WSL read as a mismatch against an identical digest — marking the host
unusable and driving the retry loop that wrote 4.9 MB of log in a day.

## Risk Assessment

Every behaviour change is behind `uses_psmux()`, so tmux and WSL paths are
byte-identical to before. Windows loses the window stamp and resolves by name
instead, which is what ADR-25 always intended there — the documented cost is
that two sessions sharing a name stay ambiguous on psmux, as `owned_target`'s
existing psmux fallback already assumed.

Not fixed here, and filed instead: psmux implements no per-*pane* user options
either, so remote hook status cannot reach thurbox from a Windows SSH host
(#1170 — `windows-vm.sh` probe A asserts the opposite). Windows installs also
never auto-update (#1172). Neither is reachable from this diff.

An operator whose psmux server already carries a global stamp needs no action:
the read-side change ignores it. Duplicate `tb-` windows left behind by earlier
relaunch cascades are genuinely ambiguous and must be killed by hand.

## Testing

`just lint`, `RUSTDOCFLAGS="-D warnings" cargo doc` and `cargo nextest run --all`
(2744 tests) all green. Seven new tests, each written to fail first:
`a_global_stamp_is_not_read_as_every_windows_identity`,
`a_global_role_never_makes_a_shell_window_an_agent`,
`only_a_multiplexer_with_window_options_is_read_as_stamping_them`,
`only_a_multiplexer_that_answers_the_attach_is_drained`,
`the_drain_stops_at_the_end_of_the_implicit_block`,
`a_psmux_shaped_stream_never_satisfies_the_drain`,
`a_coreutils_escaped_line_yields_the_digest_itself`.

Verified on real hardware — Windows 11, psmux 3.3.6, a build pinned to 2.26.8 —
against a psmux server left carrying its global stamp, so the read-side heal is
what is under test:

| | before | after |
| --- | --- | --- |
| TUI agent pane | `no live terminal` | the live PowerShell pane |
| `session capture` | `session 'alpha' has no window of its own here` | the pane's contents |
| rows' `backend_id` | `""` | `%4` / `%8` |
| windows on the server | duplicates accumulate per launch | two, one per session |

Keystrokes round-trip into the attached pane. **Not verified:** a Windows *SSH*
host (psmux over ssh) — this was a local psmux server; and the psmux quoting
paths ADR-13 already covers were not re-probed.

## Attestation

```json
{
  "head_sha": "f13e853d5b3bbb8c6630ba5b496ec3b00bc19766",
  "base": "main",
  "base_sha": "da015c975942531d943ea0768b3dfd957ab4cb69",
  "attested_at": "2026-09-17T16:10:00Z",
  "steps": [
    { "name": "review", "status": "passed", "rounds": 2, "findings": 2, "fixed": 2 },
    { "name": "lint", "status": "passed", "command": "just lint && RUSTDOCFLAGS=\"-D warnings\" cargo doc --no-deps --all-features" },
    { "name": "test", "status": "passed", "command": "cargo nextest run --all" },
    { "name": "ci", "status": "passed", "conclusion": "success", "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35245919417" }
  ],
  "verdict": "passed"
}
```

— LeTuR's agent


LeTuR (commented):
The coreutils half reproduces exactly: `sha256sum 'back\slash.txt'` prefixes the line with `\` and doubles the backslash in the path, while the digest is byte-identical to the unescaped file's — so a naive first-field parse compares `\d4e4877…` against `d4e4877…` and calls an identical file a mismatch. Confirmed here on coreutils 9.7.

The gating is on the right axis, and that is the part worth saying out loud: `uses_psmux()` is `mux() == "psmux"`, a property of the transport rather than of the build, so a Windows box driving a remote tmux over ssh keeps both the implicit-attach drain and its per-window stamps. A `cfg!(windows)` would have quietly taken those from every remote session on that machine.

I cannot check the two psmux facts underneath this — that it numbers the client's first command 1 and so emits no implicit block, and that its `@` options live in one session-wide map. There is no psmux on this machine, so those stand on #1168 and your own reading rather than on anything I confirmed.

nit: `parse_discovered(line, stamps)` reads as a bare boolean at the call site. `stamps_are_per_window()` is well named and the parameter could borrow the same word.

— LeTuR's agent

LeTuR (commented):
Retracting the caveat in my last comment: I ran both psmux facts on `windows-hp` against psmux 3.3.6, so nothing here rests on #1168 any more.

**No implicit attach block, and the client's first command is 1.** A control-mode client attached on argv opens `%window-add @1` / `%sessions-changed` / `%session-changed $64 probe` and nothing else; the first command I sent came back `%begin … 1 1`. tmux 3.5a under the same test leads with `%begin … 270 0` / `%end` *before* the session change — that is the block `drain_implicit_attach_response` consumes, and the one psmux never sends, so the `read_until` really does wait forever.

**`@` options are session-wide.** `set-option -w` of `@thurbox_session=SESS-ONE` on w1 alone made **both** windows answer `SESS-ONE`; setting w2 to `SESS-TWO` then flipped **both**. tmux under the identical commands gives w1 `SESS-ONE` and w2 empty. One session's stamp is every window's identity there, which is exactly the `Unknown`/`Absent` split behind #1168's duplicate `tb-` windows — withholding it at both ends is the right call.

The probe ran on its own `-L probe` socket and was killed after; the default socket was empty throughout. `nit:` on `parse_discovered`'s bare boolean still stands.

— LeTuR's agent

PR #1177, opened 2026-09-17
fix(core): rescan an imported folder instead of freezing it at import

LeTuR:
## Intent

A folder imported with `Alt+P` listed the repositories it held on the day you
imported it, and nothing since. Clone a repository into it and the picker never
showed it; delete one and the picker kept offering it. The only way to catch up
was to type the folder path and press `Alt+P` again — and on a remote host that
was the *only* way, because a remote folder was never rescanned at all.

## What Changed

A folder's members are a scan; the import is only the first one.

```
            ┌─────────────────────────────────────────────────┐
  before    │ Alt+P ──► repo_bookmarks (members, forever)      │
            │ local  ─ readdir ─┬─► rows   ◄── chained, so a   │
            │ remote ─ nothing ─┘           deleted repo stays │
            └─────────────────────────────────────────────────┘
            ┌─────────────────────────────────────────────────┐
  after     │ local  ─ readdir every read ──────────► rows     │
            │ remote ─ rescan_folders (30s, worker)            │
            │            └─► replace_parent_children ─► rows   │
            └─────────────────────────────────────────────────┘
```

- `flatten` takes the scan instead of a `local` flag, and a folder uses **either**
  the scan **or** the remembered members — never both chained. Chaining them was
  what kept a deleted repository on offer.
- `scan_parents` emits a folder only when `read_dir` succeeded. A key means the
  scan is authoritative; its absence means there is no scan and memory stands in.
- `RepoStore::rescan_folders` asks a remote host what each folder bookmark holds,
  per folder, on a worker, on a 30 s interval, guarded against a second
  connection while one is in flight. A scan that differs is written back with
  `replace_parent_children`; an identical one writes nothing, and neither moves
  the data epoch on its own.
- `replace_parent_children` refuses to file children under a parent that is no
  longer remembered, and `delete_repo_bookmark` removes the folder and its
  members in one transaction. A rescan is a round trip, so `d` on the folder can
  land during one, and between those two DELETEs was a state in which the
  reinsert looked legitimate.
- Left out deliberately: no rescan of a folder whose scan *failed*. The scan
  cannot tell "the folder is gone" from "the host is unreachable", so both leave
  the folder showing what it last held.

## Risk Assessment

One ssh round trip per remote folder bookmark per 30 s, only while the repo
picker is open. Off the render loop, and a scan that finds nothing deliberately
reports no data change, so it does not invalidate published groups or pure panes.

A folder that reads as **empty** is empty. That is the same answer as "you
deleted the last repository in it", which is half of what a rescan is for, but it
means a folder imported from a mount point empties while its drive is unmounted —
an unmounted mount point is a readable empty directory, and nothing in a listing
tells the two apart. It refills on the next scan after remounting.

The rescan writes to `repo_bookmarks` from a worker, where before only the
keypress-driven import did. `replace_parent_children` now reads before it writes,
so it takes the lock at `BEGIN IMMEDIATE`; there is a staged-contention test for
that.

No schema change, no config, no migration.

## Testing

Five review rounds, five findings, all fixed. Two were mine, before the first
push: a rescan reported a data change whether or not it found one (epoch churn
every interval per remote folder), and a rescan that found something dropped the
host's cached rows — which the flow reads as `loading`, so one new repository on
the host blanked the list under the cursor until the re-read landed.

Two came from the review on the first push. A rescan landing after the folder was
forgotten reinserted its children with no parent row to hold them, so `flatten`
offered each as a standalone bookmark — forgetting a folder of eight repositories
left eight loose rows. And `scan_parents` claimed an unmounted folder kept its
members, which is not what the code does; the comment and the doc now say what it
does rather than changing the behaviour to match the claim.

A fifth round, after the review said the fix above only narrowed the race: it was
right. The remaining window is *inside* `delete_repo_bookmark`, between the
DELETE of the members and the DELETE of the folder — a rescan committing there
passes the parent check, reinserts its children, and the second DELETE takes only
the folder. A 200-round racer reproduced it at round 20; both statements now go
in one transaction and it survives repeated runs.

One finding I could not evaluate: a P2, "preflight cannot prove success", whose
body was not retrievable from the API. Worth a look from a human reviewer.

CI caught one thing local lint could not: the two helpers behind the stub-ssh
test were dead code on Windows, where `clippy -D warnings` refused to build the
suite. Reproduced locally by flipping the `cfg`, fixed, and checked both ways.

Gate: `just lint` + `RUSTDOCFLAGS="-D warnings" cargo doc`, then
`cargo nextest run --all` — 2751 tests, green.

Test-first, as the repo requires. Each of these failed before its fix:
`a_folder_stops_offering_a_repository_that_has_gone`,
`a_remote_folder_is_rescanned_rather_than_frozen_at_what_was_imported`,
`a_rescan_that_lands_after_its_folder_was_forgotten_writes_nothing`,
`forgetting_a_folder_racing_its_rescan_leaves_no_orphans`.

**CI:** everything is green except `SonarQube`, which fails on
`HTTP 500 on https://sonarcloud.io/batch/project.protobuf?key=Thurbeen_thurbox&branch=main`
— the scanner cannot fetch the project's **main-branch** data. Four attempts
between 19:06Z and 19:40Z, identical every time, one of them the maintainer's own
retry. It is not this branch: #1171 failed the same way in the same window, and
the last green `SonarQube` here was #1176 at 17:35Z.

What SonarCloud's API says: the project is public, and `main` was last analysed
**2026-09-01** at commit `e7d5a7f5` — sixteen days stale, because this workflow
runs `on: pull_request` only and so never analyses `main` at all. That stale
record is what the endpoint is now failing to serve, and it was serving it
happily a couple of hours earlier. Nothing in a PR can refresh it.

`All Checks` is red only because it aggregates this one job. Greptile is green.

**Not verified:** the remote half runs against a stub `ssh` on `PATH`, not a real
sshd — it proves the scan happens, that a failure changes nothing and that the
answer is written back, but not OpenSSH itself. A real host is
`scripts/dev/e2e/linux-container.sh`'s job and was not run. Windows/psmux hosts
are untested here; they go through `scan_child_repos_script_windows`, which this
change does not touch.

## Attestation

```json
{
  "head_sha": "8333f9c28e14ebf4466fb0a142fb985011df96d2",
  "base": "main",
  "base_sha": "da015c975942531d943ea0768b3dfd957ab4cb69",
  "attested_at": "2026-09-17T19:41:38Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 5,
      "findings": 5,
      "fixed": 5
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "failed",
      "conclusion": "failure",
      "reason": "SonarQube: HTTP 500 from sonarcloud.io on /batch/project.protobuf?key=Thurbeen_thurbox&branch=main, on four attempts between 19:06Z and 19:40Z including a maintainer retry. Provider-side and repository-scoped: PR #1171 failed identically in the same window, and SonarCloud has no fresh main-branch record to serve (last analysed 2026-09-01, commit e7d5a7f5; CI runs on pull_request only, so main is never analysed). Every other check passes, Greptile included; All Checks is red only because it aggregates SonarQube.",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35262758937"
    }
  ],
  "verdict": "blocked"
}
```

— LeTuR's agent


LeTuR (commented):
`Windows (clippy)` is red on `tests/repo_memory.rs`: `stub_ssh` (489) and `one_host` (517) are defined ungated but only called from `#[cfg(unix)]` tests, so on Windows they are dead code and `-D warnings` refuses them. `#[cfg(unix)]` on both definitions clears it. Third time this shape has landed this week — after #1162's `PathBuf` and its `shell_prefix_tokens` — so it may be worth a habit of gating a helper the moment its last caller is gated.

Greptile's empty-scan P1 is real, and it lands against a sentence this PR adds. `docs/FEATURES.md` now promises "a local folder on a drive that isn't mounted, leaves the folder holding what it last held rather than emptying it", but the guard is `read_dir(..).is_ok()` and an unmounted mount point is normally an empty *readable* directory — I checked, `read_dir` succeeds and yields zero entries. So the guard passes, the scan is believed, and the folder empties. Locally that self-heals on the next read; through `rescan_folder` it is written with `replace_parent_children` and does not.

Distinguishing "read this folder and it held nothing" from "could not really read it" needs more than `is_ok()` — the cheap version is to treat an empty result as no answer for a folder that last held members, which costs you only the genuinely-emptied folder, one poll late.

nit: its other P1 reads softer than it sounds — `group_persisted` already drops a member whose folder bookmark is gone, so a late rescan cannot put rows back *under* a deleted folder. Whether they should then stand alone is a product call, not a defect.

— LeTuR's agent

LeTuR (commented):
Both are answered, and the empty-scan one better than I proposed: rather than treating an empty result as no answer, the docs now state what the code does — a readable-empty folder is empty, an unmounted mount point included — with the reason my suggestion was refused, that guessing from "it used to have members" would keep a folder you really did empty. Correcting the promise beats adding a heuristic to honour it. `Windows (clippy)` is green, so the gating took.

I under-read the other one and want to say so: I called it soft because `group_persisted` keeps rows from appearing *under* a forgotten folder, and missed that `flatten` then offers them standalone — one loose row per repository the user just forgot. That is a real consequence, and taking it was right.

The fix is airtight rather than merely present. `TransactionBehavior::Immediate` is the load-bearing part and the comment names exactly why: this reads before it writes, and a deferred transaction upgrading mid-flight returns `SQLITE_BUSY` without consulting `busy_timeout`. Check and insert are in the one transaction, so a delete committing between them cannot slip through.

nit: `replace_parent_children` now returns `Ok(())` for "wrote nothing" and for "wrote the children", and the caller reports `true` either way. Harmless today — nobody reads it for more than a warning — but a `bool` would keep the distinction the function now makes internally.

— LeTuR's agent

LeTuR (commented):
This is a race the earlier fix does not close, which is what makes it worth having found: in the window between the two DELETEs the parent genuinely *is* still remembered, so a rescan committing there passes `replace_parent_children`'s check legitimately, reinserts its children, and the second DELETE takes only the folder. Same orphans, different door.

The test earns its place, and I checked rather than assumed — its own doc says it can only under-report, so I reverted the transaction and ran it: **FAILED at round 25 of 200**, listing all eight children left behind, in 0.07s. Fast and legible for a probabilistic test, and the assertion prints the rows rather than a count.

`SonarQube` is red on infrastructure, not on this branch: the scanner died instantiating `CachedProjectRepositoriesSupplier` (`EXECUTION FAILURE`, exit 3) while SonarCloud's own quality gate for this PR reports `OK`. A re-run should clear it — there is nothing here to chase.

Clean from my side. GitHub will not let this token approve a PR it authored, so take this as the approving verdict.

— LeTuR's agent

Issue #1200, opened 2026-09-18
Shift+S crashes the sessions pane while a session is being created


LeTuR:
`Shift+S` takes the sessions pane down whenever a creation is in flight in a group that already holds a session.

```
ui/lib/order.lua:185: attempt to index a nil value (field 'session')
```

Reproduction: with `group_by_repo` on, have at least one session in a repo, press `n` and create a second session into that same repo, and press `S` while the placeholder row is still on screen.

`order.sorted_within_groups` sorts a group's root blocks by `(a[1].session.name or ""):lower()`. A `kind = "pending"` item carries `command`, not `session`, and it is its own root block — so as soon as a group holds both a session block and a placeholder block, `table.sort` invokes the comparator and indexes `.session` on the placeholder. A group with one block never invokes the comparator, which is why this only shows up once the group is not empty.

The call site takes no guard: `ui/plugins/10_sessions.lua` runs `persist_order(order.sorted_within_groups(items))` for `sessions.sort`, and unlike the move actions it does not require a selected session id, so there is nothing upstream that filters the placeholder out.

Found while reviewing #1194's branch, which does not touch this path — verified identical on `main`. The fix is presumably to sort a placeholder by something other than a name it does not have, or to exclude placeholders from the sort and re-append them, but which of those is right depends on where a placeholder should end up after a sort, so it wants a decision rather than a one-line guard.

— LeTuR's agent

PR #1202, opened 2026-09-18
feat(ui): group the session list by host when sessions span machines

### LINKED ISSUE #1194: Group the session list by host
The session list never says which machine a session runs on. A remote row gets a `⇅` mark and nothing more, so with sessions spread over several hosts the list interleaves them under repo headers and "what is running on that box" has no answer at a glance.

Measured on a live multi-host setup: 21 rows across three distinct `backend_type` values — 14 on the local backend, 5 on one SSH host, 2 on another.

```text
today                          proposed (multi-host only)
── thurbox ──────────          ── local · thurbox ──────
 ● fix-osc52                    ● fix-osc52
 ⇅ remote-build                 ● perf-cache
 ● perf-cache                  ── local · website ──────
── website ──────────           ● update-deps
 ● update-deps                 ── buildbox · thurbox ───
                                ⇅ remote-build
```

The grouping machinery already exists in `ui/lib/session_model.lua`; this is a second axis over the same rows, not new plumbing. Three things to settle before writing it.

### 1. Host nests with repo grouping, it does not replace it

The host becomes the outer axis of the existing group key: groups sort by host first (local, then remote hosts by name), and a group's header names its host and its repo set. Repo grouping keeps working inside a host, and `group_by_repo = false` still means genuinely ungrouped — one flat list per host.

Replacing repo grouping would throw away the question the list is grouped for today ("which project is this?"). A separate host header line above the repo header would mean two header levels in a model where `lib/order.lua`'s move and sort algebra reads exactly one header per group, and where a group header is glued to its group's first row — that is real structural cost for an axis that most users never see.

### 2. A machine with one host pays nothing, and it is not a setting

The axis turns on only when the live rows span more than one host. With a single host — which is every laptop with no remote sessions, and also a machine where every session is on the same remote box — the group keys, the labels and the order are byte-for-byte what they are today.

No new knob. A header that appears when there is one group is noise, and a switch defaulting to on is that same behaviour with more surface to explain.

### 3. `thurbox-cli session list` grows nothing

The command already carries the fact per row: `BACKEND` in the human table, `backend_type` in `--json`. Grouping is a view decision. Inserting header lines or resorting a stream that scripts pipe through `jq` costs those callers and gains nothing that `sort` or `jq group_by` does not already give them.

— LeTuR's agent

LeTuR:
## Intent

The operator runs sessions on more than one machine and the list never said so:
a remote row got a `⇅` mark and nothing else, so with rows spread over three
machines "what is running on that box" had no answer at a glance. Closes #1194,
which settles the two decisions this needed — whether host replaces or nests
with the repo grouping, and whether `thurbox-cli session list` grows the same
thing.

## What changed:

- The host is the **outer** grouping axis in `ui/lib/session_model.lua`: this
  machine's groups first, then each remote host's by name, `buildbox · webapp`
  in one header. Repo grouping keeps working inside a host. It is one header
  *axis*, not a second header *level* — `lib/order.lua` finds a group's edges by
  the single row that carries a header, so a second level would have to be
  taught to the move algebra, the click targets and the border dots too.
- **Derived, not a setting.** The axis turns on only when the list spans more
  than one machine. One machine renders what it rendered before — the existing
  frames in `tests/frames.rs` are untouched, which is the assertion.
- A group edge is now a host+repo edge, so a `Shift+J` that would carry a group
  onto another machine is refused. The swap would be persisted and then undone
  by the next build re-clustering the group under its own host — the shape the
  `group_by_repo` setting was reshaped to avoid.
- `Command::host()` publishes a creation's target machine to the interface, so
  a placeholder is drawn under the machine it is being created on. It answers
  in bare names via a new `session::host_name_of`, which `snapshot.rs` now
  shares: the host picker carries `ssh:devbox` while a session's machine is
  published as `devbox`, and left unreconciled the list drew a phantom
  `ssh:devbox` group beside the real one.
- `thurbox-cli session list` grows nothing. It already carries `backend_type`
  per row; grouping is a view decision, and resorting a stream that scripts pipe
  through `jq` would cost them for nothing.

## Risk Level:

Safe to merge. This pipeline runs no IaC plan.

Deliberate trade-offs a reviewer cannot see in the diff:

- **With `group_by_repo` off and several machines on screen, cross-machine
  moves are refused.** Turning repo grouping off used to mean one list you
  could reorder freely; it now means one flat list *per machine*. Ordering
  across machines has no meaning once the host clusters them, and a session
  cannot move between machines anyway — but the refusal is silent, as every
  group-edge refusal already is.
- **A creation in flight counts towards "more than one machine".** Creating your
  first session on a host is a list whose rows are all local and whose next row
  is not, so the axis flips on while it is being made. A *failed* creation keeps
  it on for the few seconds the bus holds the failure, deliberately: the
  placeholder is the only thing that says the creation happened, and it belongs
  under the machine it was asked for.
- **A host in `hosts.toml` literally named `local` renders a header identical to
  this machine's.** The keys differ (`\0local` cannot be spelled by a host name),
  so the grouping and the move refusal are correct; only the label is ambiguous.
  Pinned as such in `tests/frames.rs`.
- Headers are longer by the machine name, and `ui.rule` clips rather than
  truncates — as it already did for long repo sets.

Found beside this work and deliberately **not** fixed here: `Shift+S` crashes the
sessions pane whenever a creation is in flight in a group that already holds a
session (`ui/lib/order.lua` sorts blocks by a name a placeholder does not have).
Verified identical on `main`; filed as #1200.

## Push steps:

- **Review, four adversarial rounds**, each of which found something real.
  Round 1: a placeholder was filed under a remote host's group when that host
  was the only one holding the repo; and a host named `local` merged into this
  machine's group, silently disabling the new move refusal. Round 2: the fix for
  the first one had inverted the bug — it hardcoded "this machine", so *every*
  remote creation now lied the other way. Round 3: the published host was the
  *backend* name (`ssh:x`) while a session's is bare (`x`), so they never
  matched and a phantom group appeared; the axis gate ignored creations
  entirely; the memo digest was blind to the new field; and the `claimed` guard
  had become dead code. Round 4 found no blocking defect, and verified by fuzz
  (20k cases) and ~39k legal moves that `build` is a fixed point of
  persist→rebuild — i.e. the "accepted, persisted, then undone" failure this
  change is shaped to prevent does not occur.
- Round 3's finding also showed the test I had written could not have caught it:
  it hand-built the in-flight command with a bare host, a spelling the interface
  never sends, and asserted on line order that a phantom group still satisfied.
  It now builds the command the bus builds and asserts the phantom's *absence*.
- **Tests** were written before each change and verified to fail for the right
  reason first — including by reverting each fix in turn and confirming the new
  test failed, then restoring.
- **Docs**: `docs/FEATURES.md` (which owns feature-level design choices) and the
  `thurbox-cli` skill. Round 4 caught both still describing the old gate after
  the code started counting creations; corrected. `docs/PLUGINS.md` gained the
  newly published command field.
- **Gate** (`.publish.yaml`: `just lint`, the rustdoc check, `cargo nextest run
  --all`) green on this commit: 2798 tests, 0 failures. Earlier runs on this
  machine flaked on `tui_e2e` and on three host tests that read the real
  `hosts.toml` and see this WSL machine as a shareable host; all pass here and
  in isolation, and none is touched by this branch.
- **Not run**: `npm run lint:website`, which needs `npm ci` and is CI's own job.

— LeTuR's agent

## Attestation

```json
{
  "head_sha": "9abd942ae13b7898ce13bdd0472b501933511fc5",
  "base": "main",
  "base_sha": "985d1de3b6e9ed9f046fce22ca04da4afc2b2d13",
  "attested_at": "2026-09-18T16:46:40Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 4,
      "findings": 24,
      "fixed": 12
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35362057696"
    }
  ],
  "verdict": "passed"
}
```


LeTuR (commented):
Next: merge

Spot-checked the two places a change like this usually breaks. The `commands` group is gated on `epoch.data`, which bumps on accept and on every phase move, so the new `host` cannot go stale behind the memo; and `thurbox.commands[i].host` stops selene at the list, so no `thurbox.yml` entry is owed for it. `host_name_of` folds both prefix-strippers into one and drops the `Some("")` that `remote_host_of` used to return for a bare `ssh:`.

The four spellings are pinned by test — `ssh:`, `wsl:`, already-bare, and absent — which is the reconciliation round 3 was about, and `LOCAL_HOST_KEY = "\0local"` keeps a host actually named `local` in its own group.

nit: the gate record says "green on this commit" without naming it. The JSON attestation the other branches carry is what makes that checkable a week from now.

— LeTuR's agent

PR #1209, opened 2026-09-19
feat(ui): make grouping the session list by host a setting

LeTuR:
## Intent

#1202 made the session list group by host whenever its rows span machines, and
left no way to say otherwise: it never touched `ui/plugins/10_sessions.lua`, so
no row exists in settings for it. An operator who wants the remote sessions in
among the local ones — one list, ordered by hand across the lot — has no way to
ask for that. Grouping by repo has been a switch since it shipped; this is the
one beside it.

## What Changed

- **`sessions.group_by_host`** (settings → Sessions), **default on**, so a list
  that already draws host headers keeps drawing them.
- The axis now has two conditions instead of one: the switch, and the tally of
  machines `spans_hosts` already did. The tally is not a preference — a header
  naming the only machine on screen is noise whatever was asked for — so **one
  machine still draws no host header with the switch on**.
- With it off, `by_host` is false and everything below is built the way it was
  before the axis existed: no host in the group key, no host clustering, no
  host edge for `lib/order.lua` to refuse a `Shift+J` at. The same shape as
  `group_by_repo` off leaving no repo edges.
- The switch is in the model's memo key. It is read from the registry, which
  neither the row table nor the commands digest can see, so without it flipping
  the row in the modal is answered from the cache.
- `docs/FEATURES.md` and the `thurbox-cli` skill both said the axis was derived
  and not a setting. Both now say which half of the gate answers what.
- How grouping is computed is untouched, and so is `group_by_repo` — including
  the stored override in `registry.rs`'s decisions fixture.

**A second switch, not a choice row** — the decision the issue asked to settle.
`none / repo / host / host then repo` would read as one decision, and the
settings modal cannot render it: a plugin's declared value is `Bool`, `Number`
or `Text`, and `Text` is free text you type into (`modals/settings.rs`,
`activate`), with cycling wired to the core rows' one enum rather than declared.
That row would be a box where a misspelling silently means "none", and it needs
a choice type in the registry, in the Lua declaration and in the modal first —
a separate change. The two switches are independent axes anyway, not two
spellings of one: all four combinations already render, from one flat list to
`buildbox · webapp`.

## Risk Assessment

One new declaration and one `and` in front of an existing call. With the default
nothing renders differently, which the untouched `tests/frames.rs` literals
assert. No migration: a setting nobody has overridden resolves to its declared
default, and no existing key is renamed or removed.

## Testing

Written before the code and failing on `no setting named
"sessions.group_by_host"`: two cases in `tests/plugin_settings.rs`, beside the
`group_by_repo` ones. One repo on two machines groups by machine by default and
merges into one group with the switch off; a single-machine list renders
identically with the switch untouched, on, and off.

Gate: `just lint`, then `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps
--all-features`, then `cargo nextest run --all`. Not run: `npm run
lint:website`, which needs `npm ci` and is CI's own job.

## Attestation

```json
{
  "head_sha": "15a30a0eb0938527dcc6ade5a823ee459ec5c6d7",
  "base": "main",
  "base_sha": "64536e68dc13a6e9c3c856776508ec03777cf163",
  "attested_at": "2026-09-19T07:59:14Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 2,
      "findings": 2,
      "fixed": 2
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35430120862"
    }
  ],
  "verdict": "passed"
}
```


LeTuR (commented):
Next: merge

The memo key is the part worth checking and it is right: `host_grouping` goes into both the comparison and the write-back, and the comment names exactly why neither the rows table nor the commands digest covers it. Left out, flipping the row in the modal would have been answered from the cache — the stale-memo shape this file has been bitten by before.

`by_host = host_grouping and spans_hosts(...)` keeps the tally as the half that is not a preference, so one machine still draws no header with the switch on. That is what lets the untouched `tests/frames.rs` literals stand as the no-visible-change assertion.

Both places claiming the axis was "derived, not a setting" are gone — `docs/FEATURES.md` and the `spans_hosts` docstring itself — so nothing is left contradicting the switch.

— LeTuR's agent

Issue #1193, opened 2026-09-18
advertise_running_cli links thurbox-cli to itself, and nothing ever repairs it


LeTuR:
`advertise_running_cli` links `thurbox-cli` **to its own path** when the CLI it resolves already lives in the directory it advertises into. The link is self-referential, every use of it is `ELOOP`, and nothing ever repairs it.

Measured inside an auto-discovered WSL distro on 2.27.6, after #1182's fix let the mirror get this far:

```
$ ls -la /home/thurbox/.local/share/thurbox/bin
-rwxr-xr-x 11867536 Sep 18 09:44  thurbox
-rw-r--r--     1065 Sep 18 09:44  LICENSE
lrwxrwxrwx       50 Sep 18 11:01  thurbox-cli -> /home/thurbox/.local/share/thurbox/bin/thurbox-cli

$ readlink -f .../bin/thurbox-cli   # rc=1, no output
$ .../bin/thurbox-cli --version
sh: Too many levels of symbolic links
```

Note the timestamps: `thurbox` and `LICENSE` were extracted at 09:44; the self-link appeared at 11:01, from a later pass. The release tarball ships `thurbox-cli` as a **real 12.5 MB file**, so nothing was missing — the link was written over a path that should have held it.

## Mechanism

`advertise_running_cli` (`src/session_ops/host_cli.rs:208`) guards two things and not the one that matters:

```rust
let target = crate::agent::tmux::resolve_cli_binary();
if !target.is_absolute() || !target.exists() { return; }   // no `target != link`
let link = dir.join("thurbox-cli");
```

`dir` is `<data dir>/bin`. On a provisioned host the CLI *is* `<data dir>/bin/thurbox-cli`, so `target == link` and the `symlink(&target, &link)` below points the path at itself.

**It is permanent once written**, which is the part that turns a slip into a lasting break — both exits repair nothing:

- `read_link(&link).is_ok_and(|current| current == target)` reads the self-link, finds it equal to `target`, and returns early as though it were correct;
- and once `target` is the loop, `target.exists()` is false, so the next call returns at the guard instead.

## What it costs

```mermaid
flowchart TD
  A["advertise_running_cli:<br/>target == link, no guard"] --> B["thurbox-cli -> itself (ELOOP)"]
  B --> C["host CLI probe fails"]
  C --> D["mirror of wsl:Ubuntu skipped"]
  D --> E["socket never learned"]
[...]

PR #1210, opened 2026-09-19
fix(core): verify a provisioned thurbox-cli and let a broken one be replaced

LeTuR:
## Intent

A host was handed a `thurbox-cli` that was **54% of itself** — 6,815,232 bytes,
its ELF header still declaring section headers at 12,627,792, executable,
segfaulting on every call — and stayed unusable for about a day. The corruption
was transient; its effect was not. Removing the file and re-probing produced a
correct 12,640,144-byte binary that ran on the first try, on the first attempt,
with nothing about the host or the network having changed. The only thing
keeping the machine broken was the bad file nobody checked.

Three defects sat behind that, and the third is the one that mattered.

**Nothing verified the binary after it was installed.** `provision` fetched the
archive, shipped it, extracted it, logged `provisioned thurbox-cli <version>`
and returned `Ok`. `fetch_archive` checksums the download on *this* machine;
nothing checksummed what landed on the host, and nothing asked the installed
file whether it ran.

**The failure was misdiagnosed.** The probe script ran the CLI as its own last
command, so a segfault arrived here as no output at all and was reported as
`printed no JSON version (EOF while parsing a value at line 1 column 0)` —
which sends an operator to look at versions and protocols for a file that is
half a binary.

**And the broken artifact blocked its own repair.** `provision` is reached
through the mirror pass, and that pass is skipped precisely because the probe
failed:

```mermaid
flowchart TD
  A["corrupt thurbox-cli installed"] --> B["probe runs it: segfault, no output"]
  B --> C["host marked not usable"]
  C --> D["mirror pass skipped"]
  D --> E["provision never runs again"]
  E --> A
  C --> F["backoff grows 60s -> 120s -> 240s"]
  F --> B
```

The one mechanism that would have replaced the bad binary was disabled by the
bad binary, and the backoff made it quieter over time without ever making it
better.

## What Changed

- **`provision` asks the binary it installed for its version, and returns what
  it answered with** (`CliInfo`, not a path). A `provisioned` log line is now
  written only after the thing being announced has spoken. `establish` no
  longer probes the fresh copy a second time — the verification *is* that
  probe, so the round trip it used to cost is gone rather than doubled.

- **The probe's line protocol carries `@status <n>`**, what the binary it found
  exited with, between `@cli <path>` and the JSON. It is the whole reason a
  signal death can be told from a version mismatch: a POSIX shell reports
  `SIGSEGV` as 139, so the message is now `thurbox-cli at <path> died on signal
  11 (SIGSEGV) instead of reporting a version`. Four outcomes are now four
  messages — no CLI at all, a death on a signal or a crash, a clean exit that
  printed nothing, and output that would not parse.

- **Where the deadlock was broken, and why there.** `probe` returns a
  `ProbeFailure` that says whether the host *answered*, and `establish` treats
  the two as the different states they are: a host that did not answer returns
  `No` and keeps its backoff, while a host that answered with a broken CLI
  falls through to provisioning exactly as a host with no CLI does.

  It belongs there and not in the mirror because the mirror cannot tell the two
  apart — `usable()` hands it one `No` and a reason string, and branching on
  what that string says is exactly the mistake `Reach` exists in this file to
  avoid. The probe is the only layer that knows the host ran the script and the
  binary it found is broken, so the probe is where that fact is kept. Nothing in
  `kernel::terminal` changes: its 60-second retry after a pass that could not
  run is correct and still applies, and it now gets a usable verdict because the
  host was repaired rather than because the pass was retried.

- **Left out on purpose:** no checksum of what lands on the host. See the risk
  section — the measured file was not written by the transfer this change could
  have checksummed.

- Docs: ADR-24 in `docs/ARCHITECTURE.md`, the provisioning and retry bullets in
  `docs/FEATURES.md`, and the `thurbox-remote-hosts` skill.

## Risk Assessment

**Windows and WSL are unverified.** The failure was measured on a Windows
machine's WSL distro, and everything here is exercised on Linux against a
planted file. The Windows halves — `$LASTEXITCODE` carrying an NTSTATUS through
the PowerShell probe, `Expand-Archive` — are written from the documented
behaviour and reviewed, not run. They want checking on real hardware.

**A host whose CLI is broken now costs a provisioning attempt where it used to
cost nothing.** That is the point of the change, and it is bounded the same way
every other provisioning failure is: if the replacement also fails to answer,
`establish` returns `No` and `retry_after` backs the host off toward the
15-minute ceiling. A host that is merely unreachable is untouched and has a test
of its own saying so.

**The probe's line protocol changed**, but the script and the parser ship in the
same binary and the script is generated here, so there is no version skew to
carry. A missing `@status` line is read as a clean exit rather than as a
failure.

**On what wrote the corrupt file.** `advertise_running_cli` is not the writer:
it only ever creates a symlink, never a regular file, and since #1193 it returns
without touching a regular file at that path — the artifact was a 6,815,232-byte
regular executable. The evidence that pointed away from a `tar` extraction was
the mtimes: `LICENSE` and `thurbox` read `09:44` while `thurbox-cli` read
`11:49`, two hours later. That turns out to be what a **cut-short extraction**
looks like, reproduced here with GNU tar against a truncated `.tar.gz` whose
members carried a `09:44` stamp:

```
LICENSE      1000     09:44        -rw-rw-r--
thurbox      3000000  09:44        -rwxr-xr-x
thurbox-cli  2399232  <wall clock> -rwxr-xr-x   <- cut short
tar: Unexpected EOF in archive        rc=2
```

tar takes a member's mtime from the archive header only once it has finished
writing that member, and takes the mode when it creates the file — so the
completed members keep the archive's own stamps, and the interrupted one keeps
the wall clock of the write and is still executable. That is exactly the
measured signature. So the likeliest writer is the host's own `tar -xzf`, in a
re-provisioning that failed partway; a direct `ship()` of the binary on the
dev-build path would look the same, and **I could not determine which from
here.** That ambiguity is the argument for the shape of this fix: a checksum
after the copy would have passed in the first case and proved nothing about the
file that was actually there, and anything outside this repository writing that
path would look identical. The question worth asking is not "did the bytes
arrive" but "does the thing there work", asked of whatever is there now.

It also explains why nobody saw the extraction error again. That failed pass
*did* report it — and left the half-binary behind, after which the probe failed
first on every later pass, so provisioning was never reached to fail again.

## Testing

The test was written first and failed on `main` for the right reason: `provision`
returned `Ok("…/thurbox-cli")` for a planted binary that segfaults, and the probe
reported that binary as `printed no JSON version (EOF while parsing a value at
line 1 column 0)` — the production message, reproduced verbatim on Linux.

Four tests drive the real code against a real `/bin/sh` and a file on disk, so
the shell script under test is the shell script that ships:

- `provisioning_a_binary_that_does_not_run_is_not_a_success`
- `a_host_whose_cli_segfaults_is_reprovisioned_rather_than_locked_out` — the
  deadlock end to end: the corrupt file is planted where the probe looks, and
  the host must come back usable with the binary replaced
- `a_host_that_does_not_answer_is_not_re_provisioned` — the refusal that had to
  survive the change
- `the_probe_names_a_signal_death_rather_than_calling_it_bad_json`, plus
  `a_broken_host_cli_is_named_by_what_it_did` over the parser directly

Three review rounds, five findings, all fixed: a Windows NTSTATUS printed
unsigned read as `exited 3221225477`; `verify_provisioned` wrapping every probe
failure as "the provisioned …" so a transport failure read as
`the provisioned could not start thurbox-cli probe on 'x'`; two test-only
setters left as dead code under the Windows clippy job; and two coverage gaps
(the unreachable-host refusal, and asserting the planted file is actually
replaced rather than inferring it from the verdict).

Gate: `just lint` and `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps
--all-features` clean, `cargo nextest run --all` 2835 passed / 0 failed. The
documentation step updated ADR-24, the two `docs/FEATURES.md` bullets and the
`thurbox-remote-hosts` skill in the first commit.

## Attestation

```json
{
  "head_sha": "e3a8c9a393d69d2fc3b9a1b1ed1ded75c54e4c8e",
  "base": "main",
  "base_sha": "38984ae641e5ff8d42c9af99761040a9df135830",
  "attested_at": "2026-09-19T08:35:16Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 3,
      "findings": 5,
      "fixed": 5
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35431681866"
    }
  ],
  "verdict": "passed"
}
```


LeTuR (commented):
Next: merge

The deadlock argument holds where it matters: `establish` now splits "did not answer" from "answered with a broken CLI", and only the first keeps its backoff. That is the right layer — `usable()` hands the mirror one `No` and a string, and the probe is the only place that knows the host ran the script.

I checked the thing this change could have undone and it did not: `advertise_cli_in`, `heal_self_link`, `same_path` and `link_cli` are untouched, so #1193's rule that a regular file at that path is somebody else's still stands. Repair reaches only a binary the probe watched fail. `verify_provisioned` asking the installed file rather than checksumming the transfer is also the right question given the two-hour mtime gap.

A missing `@status` reads as a clean exit and its line stays in the body, so the parser degrades to what it did before rather than failing shut.

Unverified and worth a pass on real hardware when one is free: the PowerShell half, where `$LASTEXITCODE` carries the NTSTATUS. You named it in the risk section, and this machine has no Windows host to settle it on.

— LeTuR's agent

Issue #1211, opened 2026-09-19
Session list selection follows a row index, so a session opening or closing above the cursor steals the highlight and the agent pane


spscream:
### Why

With the cursor on a live session, creating or deleting another session whose row lands *above* that cursor moves both the list highlight and the agent pane onto a different session. Keyboard focus stays where it was. The pane comment says the cursor follows the session id; on an ordinary rebuild it follows the old index.

Reproduced on current `main` (`a712d7d8`) against the real binary (3/3 opens, 3/3 closes) and against a painted `sessions` + `agent` frame (3/3 insert-above, 3/3 delete-above). Inserting *below* the cursor does not jump (3/3).

This is not #1200 (Shift+S nil `.session` on a pending row) and not #1192 (self-heal spawning a second session). Those can change the list; they are not this cursor.

### Steps

Default `group_by_repo` (on). Three sessions in three repositories whose names sort `aaa`, `bbb`, `ccc`. The agent pane is focused.

1. Select the last session (`keep-three`) with `Ctrl+J`.
2. From another process: `thurbox-cli session create --name new-zero --repo-path <repo-that-sorts-first> …`
3. The new group appears at the top of the list. The agent title that read `keep-three (…)` now reads `keep-two (…)`. The highlight is on `keep-two`. The action band still says `Agent`.

Same setup, select the middle session (`keep-two`), then `thurbox-cli session delete --force` the first. The agent title becomes `keep-three`. The band still says `Agent`.

A create or delete that only appends *below* the selected row leaves the highlight and the title alone.

### Path

`ui.cursor` (`ui/lib/ui.lua`) restores `state["sessions.cursor"]` as an index. It remaps that index onto an id only when a follow, a foreign `store.selected` write, or `store.focus_session` is set. A snapshot rebuild with neither of those leaves the index in place and `_settle` publishes whichever session now sits there to `store.selected`.

The sessions pane comment in `ui/plugins/10_sessions.lua` (`The cursor follows the SESSION it was on rather than a row number`) describes the follow path, not this one.

`ui/plugins/20_agent.lua` draws whichever id is in `store.selected`. There is no second focus state for “which session is on screen”. `Command::Focus` is not issued, so the keyboard does not move.

`session.post_create` / `focus_new_session` do not run for a create that this interface did not perform (`thurbox-cli session create`, dispatch, a second instance). Existing tests only cover that event and a steer at a missing id.

### What not to do

[...]

PR #1215, opened 2026-09-19
fix(ui): follow the selected session rather than its row number

LeTuR:
## Intent

With the cursor on a live session, another session opening or closing *above* it
moved both the list highlight and the agent pane onto a different session, while
the keyboard stayed where it was — so the operator was reading one agent and was
silently handed another. Under a fleet, which opens and closes sessions
constantly and without warning, this fires all the time. Issue #1211.

## What Changed

- **`ui.cursor` remembers the selected item, not its row.** `_settle` already
  wrote down the row number; it now writes down the identity under the cursor as
  well, and every build re-derives the row from that identity. The row number
  stopped being the first answer and became the last one.
- **Resolution order is now explicit**: the row being chased (a follow, a
  foreign `store.selected` write, a `focus_session` request), then the row the
  cursor was already on, then the remembered row number. Previously only the
  first of those ever remapped an index onto an id, so a plain rebuild — the
  common case, and the only case a `thurbox-cli session create` or a second
  instance produces — kept the stale number.
- **`select_by_id` folds into the same `row_of` lookup** rather than carrying its
  own scan.
- **When the selected session has gone**, the cursor lands on whatever now
  occupies that position: the neighbour below it, or the list's last row once the
  list is shorter than the cursor. Not the top, which is the same theft in
  another direction, and not nothing, which would blank the agent pane over a
  session the operator never closed. It is also what the model already assumed —
  `session_model.build` drops a dead row before anything is grouped or ordered
  precisely so that "the cursor lands on the next row". A list that empties
  completely publishes nothing, and the agent pane falls back to its own
  *No Session* frame.
- **The sessions pane's comment claimed this behaviour before the code had it**
  ("the cursor follows the SESSION it was on rather than a row number"). It now
  says what runs, and records the gap, because a comment describing the intent
  instead of the code is how this survived to be reported from outside.
- The change is in `ui/lib/ui.lua`, so it reaches the restore float's list too —
  same bug class, same fix, and `ui.reset` clears the new state alongside the
  old.

Deliberately not done: nothing hangs off `session.post_create` or
`focus_new_session`, and nothing subscribes to `session.created`. A CLI create
never fires the first, the setting would move you *onto* the new row rather than
leave you where you were, and the third is the theft this list already refuses.

## Risk Assessment

- **One behavioural change beyond the bug**, and it is the one the fix is made
  of: a list reordered under the cursor now keeps the selection instead of
  keeping the row. Every pane built on `ui.cursor` gets it — the session list and
  the restore float are the two.
- **A stale `follow` is not cleared, and this PR does not change that.** A follow
  is dropped when you move the cursor yourself but not when a build lands on it,
  so it outlives its own success. If a steered-to session is then deleted and
  later restored, the surviving follow pulls the cursor back onto it. That is
  pre-existing — the old code resolved a follow the same way — and fixing it
  would change `follow`'s contract for both panes, which is a second change and
  not this one. Raised here so it is visible rather than shipped silently.
- **No migration, no data effect, no ordering requirement.** The new value lives
  in a plugin's `state`, which survives a reload and not a restart, so a stale
  one cannot outlive the process that wrote it.
- **Cost per build is one extra linear scan of the item list** when nothing is
  being chased, over tens of rows. Writing the same value to `state` every frame
  does not bump the change counter (`api.rs` only counts a value that moved), so
  this does not make the screen dirty. No performance claim is made and none is
  measured.

## Testing

Written test-first: the two tests for the bug failed on `main` for the right
reason — selection `aaa` where `bbb` was held on an insert above, `ccc` on a
delete above — and the removal tests passed there already, which is what makes
them a pin on documented behaviour rather than on the fix.

New in `tests/session_list.rs`, the file that owns this surface (`tests/pane/`
does not exist in this repository, and `docs/REVIEW.md` asks for coverage in the
file that owns the surface rather than a new one beside it):

- a session opening above the cursor moves neither the bar nor the agent pane,
- a session closing above it does not either,
- the selected session going away hands the selection to the row below it,
- the last session going away clamps to the new last row,
- the last session of all going away publishes nothing.

The assertion is the painted frame, not just `store.selected`: the list
published the right value after a steer and after a create all along, so a test
that only reads the published value stays green whether or not the remap is
wired. Each of the first three checks all three readings together — the value
published, the row wearing the selection bar, and the agent pane's title — since
one write moved all three; the first also asserts no `Focus` command was issued,
because the complaint is that the keyboard and the highlight diverge.

Review: 4 rounds, 4 findings, 3 fixed. Three came from the local rounds — the
missing coverage for a list that empties, a comment on that new test claiming to
guard more than it does, and a placeholder local name clippy rejects. The fourth
was Greptile's, and it was right: the `paint` helper's doc opened by narrating
the code it sits above, which this repository's comment rules cut, so it now
carries only the reason the helper exists. The one left open is the stale
`follow` above, with its reason. Gate: `just lint` plus the rustdoc check, then the suite — 2840 tests,
all passing. The suite was run as two complementary halves (`kind(lib)` then
`not kind(lib)`) because the single run is killed by memory pressure on this
machine; together they are the same set as `cargo nextest run --all`.

CI is green on this head: 17 checks passed — including `All Checks`,
`Windows (tests)`, `Greptile Review` and the pipeline's own re-run of the same
gate — and the six that report `skipping` are the path-filtered jobs this diff
does not touch (install scripts, ShellCheck, website lint, winget packaging).
Greptile's one finding is the fourth counted above and is now resolved.

Not verified locally: the website linters, which `.publish.yaml` leaves to CI's
own job, and which CI skipped here because nothing under `website/` changed.

The base branch advanced by one commit (#1212, a GitHub Action version bump)
after the gate ran. The branch is not rebased onto it: the pull request is
mergeable, that commit cannot affect a Lua and test change, and rebasing would
discard the pipeline run this attestation is about. `base_sha` below therefore
names `eedc421a`, which is the commit this branch is rebased onto and the base
the reviewed diff reconstructs against.

## Attestation

```json
{
  "head_sha": "b3de9255c58acf2a59cd5e8540c2d84a5c48d4e6",
  "base": "main",
  "base_sha": "eedc421af479dd8bea60f79b693e239e54b21a1a",
  "attested_at": "2026-09-19T15:07:46Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 4,
      "findings": 4,
      "fixed": 3
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
      "command": "cargo nextest run --all -E 'kind(lib)' && cargo nextest run --all -E 'not kind(lib)'"
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35448893691"
    }
  ],
  "verdict": "passed"
}
```

https://claude.ai/code/session_01Segh8VQiy6jU9cZiLpJbRf


LeTuR (commented):
Next: @LeTuR — the code is right; the one finding is cosmetic and yours to settle.

The resolution order does what the body claims: `row_of(follow) or row_of(selected)`, with the remembered number reaching `_settle`'s clamp only once the selected item has gone. That last part is the half that is easy to get wrong, and landing on whatever now occupies the position — rather than the top, or nothing — is the right answer for the same reason the model already drops a dead row before grouping.

Putting it in `ui.lua` rather than `10_sessions.lua` is what makes the restore float inherit it, and `ui.reset` clearing `.selected` beside `.cursor` is what stops a pane reload resurrecting a stale identity.

Greptile's only finding is the test helper's doc narrating its steps. I would take it rather than argue it — it is two lines, and a red check is how this repo reads "not done".

— LeTuR's agent

LeTuR (commented):
Next: merge

The delta since `17c8cf4b` is one doc comment, now carrying why the painted frame is the only evidence that catches this class — `store.selected` was already right after a steer and after a create, so a test reading it alone stays green either way. That is the finding taken rather than worked around.

I checked the load-bearing perf claim instead of taking it: `state`'s `__newindex` (`api.rs:505-522`) bumps the version only when a value actually moved, so writing `.selected` on every settle does not make the screen dirty.

Leaving the stale `follow` is the right call — pre-existing, and clearing it would change `follow`'s contract for both panes built on `ui.cursor`.

The attestation reads `verdict: blocked` only because it was written before the pipeline reported. It names this head, every local step passed, and CI is green on it now.

— LeTuR's agent

PR #1219, opened 2026-09-21
fix(core): let Windows installs update themselves

### LINKED ISSUE #1172: Windows installs never auto-update: the updater refuses and the default-on setting says nothing

LeTuR:
## Intent

`auto_update` is on by default and has never done anything on native Windows:
the updater refused outright, and because the startup check is a background task
it refused silently. Closes #1172.

The refusal was about artifact shape, not policy. Making it work took four
changes, one of them the real work.

## What Changed

1. **The zip is unpacked.** The unpacker is chosen from the archive's own
   extension, not the running platform, because `fetch_archive` fetches a
   *peer host's* artifact. A `.zip` goes to PowerShell's `Expand-Archive` —
   what `install.ps1` uses, and the only unpacker a native Windows box is
   guaranteed to have. No new crate dependency.
2. **A running `thurbox.exe` is replaced, with `ReplaceFile` instead of a
   rename.** See below.
3. **The binaries carry `.exe`.** `install_binaries` now appends the platform's
   executable suffix; without it the zip path reported the archive as missing
   `thurbox`.
4. **The triple guard names Windows**, not `-windows-msvc`, so a `-windows-gnu`
   target stops falling through into the tar path and asking the release for an
   artifact nothing builds.

### The swap: `ReplaceFile`, not rename-aside and not an after-exit helper

`std::fs::rename` is `MoveFileEx(REPLACE_EXISTING)`, which has to *delete* the
destination, and Windows will not delete an image a process has mapped. Win32
`ReplaceFile` does not delete it: it renames it to a backup name, which a mapped
image permits. So the swap succeeds **while thurbox runs**, and the running
process keeps its old image until the next launch — the same thing Unix does
with the inode. It is reached through PowerShell's `[System.IO.File]::Replace`
rather than a hand-written Win32 binding, since the zip path already needs
PowerShell.

The two approaches the issue names both lost to it:

- **Rename-aside** (move the running exe away, move the new one in) has a window
  between the two renames in which the name `thurbox.exe` resolves to nothing.
  Reproduced on real Windows: after the first rename, `Test-Path thurbox.exe`
  is false.
- **A helper that swaps after exit** works, but it adds a detached process, a
  wait keyed on a PID (which is reusable), and a timeout, all for the same
  single replace `ReplaceFile` performs anyway. I built it first and dropped it
  once `ReplaceFile` turned out to work on a running image.

**What an interrupted swap leaves:**

```mermaid
flowchart LR
  A[download + verify + stage<br/>.thurbox.exe.new] -->|interrupted| OLD1[thurbox.exe = old]
  A --> B[ReplaceFile<br/>one system call]
  B -->|thurbox killed / closed| OLD2[old or new, never neither]
  B --> NEW[thurbox.exe = new<br/>.thurbox.exe.old = old image]
```

- Before the swap, nothing installed has been touched.
- The swap is one system call. A thurbox that is killed, crashes or is closed
  cannot stop the kernel halfway through it, so `thurbox.exe` is the old binary
  or the new one.
- The old image is kept as `.thurbox.exe.old` in every case. The next update
  removes it once nothing runs from it.
- Updating twice without restarting — the second time while a thurbox still runs
  from the backup — is refused by `ReplaceFile` **before** it touches
  `thurbox.exe`, and reported as a failure. Verified: the installed binary stayed
  the previous good one, with the staged file and the backup intact.

## Risk Assessment

**The Unix path is unchanged**: the same rename and the same error message, byte
for byte. The difference is confined to the `#[cfg(windows)]` `commit_binary`.

**Known limit, deliberately left.** `thurbox-cli update` run while a TUI that
has *already* self-updated is still open is refused (see the last bullet above),
and the error now says the binary may be in use. Working around it needs a
unique backup name per run and a sweep to clean them up. Not worth it for a
case that fixes itself on the TUI's next launch.

**Power loss inside the call is not covered by testing.** `ReplaceFile` is not
documented as crash-atomic against an OS crash or power cut in the middle of
the call. What holds there is that the old image is never deleted, so recovery
is a rename, not a reinstall.

**Out of scope:** `install.ps1` has the same locked-exe problem on an in-place
reinstall.

## Testing

Review took two rounds. Round one found one defect, now fixed in its own
commit: `Expand-Archive` shares thurbox's console, and the startup update runs
while the interface owns the screen, so its progress bar would be drawn over the
frame. `tar` prints nothing on success, which is why the Unix path never had
this. It is silenced with `$ProgressPreference`. That one is reasoned from the
progress records `Expand-Archive` emits, not seen on a screen. Round two was
clean.

Gate on the attested head: `just lint`, `cargo doc` with `-D warnings`, and
`cargo nextest run --all` (1966 + 894 passed). Also clean: `cargo clippy
--target x86_64-pc-windows-gnu --all-targets -D warnings`.

Four new tests, all failing on `main`:

- `archive_name_names_windows_not_one_triple` — the guard. On `main` it fails as
  an assertion, not a compile error:
  `left: "…-windows-gnu.tar.gz"  right: "…-windows-gnu.zip"`
- `extractor_for_picks_expand_archive_for_a_zip` — the decode
- `install_binaries_appends_the_platform_exe_suffix` — the `.exe` suffix. It is
  driven by the suffix, so it runs on Linux too; on Windows it goes through the
  real `ReplaceFile` path and asserts the backup, so the `windows-test` job
  exercises the swap natively.
- `replace_file_script_keeps_the_old_image_as_a_backup` — the exact PowerShell
  the swap runs

**On a real Windows 11 machine (PowerShell 5.1)**, not by reasoning:

| what | result |
|---|---|
| `MoveFileEx(REPLACE_EXISTING)` over a running image — what `std::fs::rename` does | refused: *Access is denied* |
| `ReplaceFile` with a backup name, over a running image | **replaced**; the process kept running |
| `ReplaceFile` with no backup | refused — the backup is why it works |
| the generated swap script, as a child `powershell.exe`, refused | exit 1 |
| the generated `Expand-Archive`, on the latest real release zip | flat `thurbox.exe`, `thurbox-cli.exe`, `LICENSE` |
| this branch's `self_update` tests, cross-built and run natively | 15 passed |
| **end to end:** a cross-built `thurbox-cli.exe` from this branch in a throwaway install, `update --force` | exit 0; `0.0.0-dev` → `2.29.2`; it replaced **itself while running**, `thurbox.exe` too, both old images kept as `.old` |

The end-to-end run used an isolated config dir, data dir and mux socket, and
touched nothing outside its temp directory. The Windows runs were made on the
commit before the review's progress-bar fix; that fix adds one PowerShell
preference to the extract command, run on the same machine separately.

**Not verified:** a power loss or OS crash in the middle of the `ReplaceFile`
call; and the progress bar over a live TUI (see above). An operator check on a
real install is still worth doing: `thurbox-cli update` from the first release carrying this,
then `thurbox-cli version`, and check `.thurbox.exe.old` is cleared by the next
update.

## Attestation

```json
{
  "head_sha": "7c7ef6104356bb141503dce91a05fac8b22e6bd8",
  "base": "main",
  "base_sha": "1c3cec7bac2ddad0c60c7dc6f8c867505ad1205d",
  "attested_at": "2026-09-21T09:07:09Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 2,
      "findings": 1,
      "fixed": 1
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
      "command": "cargo nextest run --all -E 'kind(lib)' && cargo nextest run --all -E 'not kind(lib)'"
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35580077034"
    }
  ],
  "verdict": "passed"
}
```

— LeTuR's agent


LeTuR (commented):
Next: merge

The Windows claim is proven where it matters, not reasoned: real Windows 11 on PowerShell 5.1, and `install_binaries_appends_the_platform_exe_suffix` takes the real `ReplaceFile` path under the `windows-test` job, so CI exercises the swap natively on every run. Choosing `ReplaceFile` over rename-aside is the right call for the reason given — the window where `thurbox.exe` resolves to nothing is reproduced, not hypothetical.

I checked the part that is easy to get wrong in generated PowerShell: every path goes through `shell::powershell_quote`, which doubles a single quote, so a profile directory with an apostrophe in it survives; `-LiteralPath` keeps a `[` from being read as a wildcard; and `File.Replace` gets staged → installed → backup in the order .NET expects.

The one piece reasoned rather than seen is `$ProgressPreference` keeping the `Expand-Archive` progress bar off the interface — you said so, and it is the right thing to have said.

— LeTuR's agent

Issue #1220, opened 2026-09-21
Updating thurbox ui went wrong, shell under session list


rchoffar:
I typed
```
I want to update my ui⡀of thurbox to have the shell of the session bellow the multiple session⠁list, but make it a little more large
```

In the ui plugin and the agent shell went flashing constantly

First codex analyse : 
```
I found the cause: thurbox assumes the agent and shell share one tab area. Showing both at different sizes made them repeatedly resize each other, causing
  the flashing.
```

LeTuR:
Second reporter, and the symptom is wider than flashing: **opening the shell pane changes the agent pane, and the shell does not render properly in its own pane.**

Layout evidence for the shared-tab-area theory. In a working `layout.lua`, `shell` is a slot in the LEFT column under `sessions` (`pct = 35, min = 6`) while `center` is the agent — two separate areas, no shared tab. That file's own comment records the transition: *"The shell was a second TAB of the centre until it became a pane, which is a change only this file and `30_shell.lua` had to know about"*. If only those two knew, whatever still treats the two as one tab area never learned, which is what makes them resize each other.

nit: the flashing is the visible half. The shell not drawing its own content is the half that suggests the state, not just the geometry, is still centre-owned.

— LeTuR's agent

LeTuR:
Third symptom from the same reporter, and it tells the two theories apart: **the agent pane takes the same SIZE as the shell pane.**

In the layout that triggers this, those two are nowhere near the same size — `shell` sits in the left column at `pct = 35, min = 6`, while the agent holds `center`, the whole remaining width. An agent rendered at the shell's dimensions is one geometry computed once and applied to both, which is a shared tab area doing what a shared tab area does.

That is not a resize feedback loop with the panes converging on a bad size — it is the same size, deliberately. The flashing is what happens when that fights the layout's own arithmetic on the next frame.

— LeTuR's agent

LeTuR:
<img width="688" height="554" alt="Image" src="https://github.com/user-attachments/assets/dcff24c1-5016-46c9-ba6e-269e6f78f981" />

Should be ok now!

PR #1222, opened 2026-09-21
fix(core): give the agent pane and its companion shell each their own geometry and state

LeTuR:
## Intent

[#1220](https://github.com/Thurbeen/thurbox/issues/1220): with the companion
shell and the agent on screen at once — an operator's `layout.lua` put the shell
in the left column under the session list and left the agent in `center` — the
agent renders at the shell's size, opening the shell reflows the agent, the
shell draws its own screen at the agent's width, and the two repaint each other
continuously.

Four symptoms, one mechanism. `layout.lua` is the operator's file, so the bug is
that a legal arrangement broke the panes.

## What Changed

`kernel::terminal::Live` held **one** rect, **one** size memo and **one**
"which view is in it" flag per session, and `Session::resize` sized both panes
together. That is the model from when the shell was a second tab of the centre,
where only one of the two could ever be on screen. Give the shell a slot of its
own and each frame wrote its own rect into the shared memo and resized *both*
panes to it; the other pane spent the next frame undoing that.

```mermaid
flowchart LR
  subgraph before["before — one geometry, two panes"]
    A1[agent paints<br/>90x38] --> M[(Live.size<br/>Live.rect)]
    S1[shell paints<br/>30x13] --> M
    M --> R1[Session::resize<br/>sizes BOTH panes]
  end
  subgraph after["after — one geometry each"]
    A2[agent paints<br/>90x38] --> MA[(agent: Painted)] --> RA[resize]
    S2[shell paints<br/>30x13] --> MS[(shell: Painted)] --> RS[resize_shell]
  end
```

`Painted { size, rect }` is one record per painted surface — the agent's, its
`#shell` companion's, and a plugin's program pane, which already worked this
way. A `Pane` resolves a surface *name* to one of a session's two panes once,
and carries the answers, so nothing threads "which of the two" alongside the
session. `render_session` is one path for either pane instead of two that shared
cells. Everything keyed on "the session's surface" — the rect a click resolves
against, the pane a wheel tick or a drag reaches, the grid a copy, a link scan
or a search reads — is keyed on the surface name.

`ui/layout.lua` is untouched.

## Risk Assessment

**A plugin that asks about a session by its bare id while rendering the shell
now gets the agent's answer** from `visible_text` and `links`. It should ask
about `<id>#shell` — the name it already spells to render it. No bundled or
example pane issues `copy`, and `thurbox.cmd.Session` documents the field as
taking a surface.

**Two deliberate trade-offs**, both raised in review and both kept:

- *A shell pane is born at the terminal's size, not at the agent's rect.* The
  old birth-at-rect clamp would now be safe (the memos are per-surface, so any
  real rect still differs from an empty one) and it did fit the tabbed default,
  where the agent's rect **is** the shell's slot. It is gone anyway, because
  deriving one pane's dimensions from the other's is the thing this change
  removes. The cost is one reflow on a freshly-opened shell in the tabbed
  arrangement, corrected by the first frame that paints it.
- *`screens()` reports both of a session's panes under that session's id.*
  Search asks which session holds the text, and the shell is a screen of that
  session. Before, it scanned whichever pane happened to be painted, so the
  answer depended on the arrangement. The limitation this leaves: the search
  strip selects a session, not a surface, so a hit that is only on the shell can
  send you to a session where it is not the visible pane. Making search
  surface-aware is a separate change.

Reported, not fixed: `docs/FEATURES.md:1155` and `:1248` reference
`TerminalView`, which no longer exists in v2. They predate this branch.

## Testing

`just lint`, `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features`
and `cargo nextest run --all`.

New: `src/kernel/terminal/decoupling.rs` — 12 tests driving the real
`layout::resolve` and `paint::render` against a backend that records the sizes a
multiplexer would be sent. Four arrangements: the reported one, each pane alone,
and the inverted one (agent in the left column, shell in `center`), which is
what catches anything still assuming the agent owns the centre.

Run against the pre-fix kernel — the file adapted to the old API shape, which is
`Live`'s field list and three call sites — six of them fail, and they fail with
the reported symptoms rather than incidentally:

```
the shell's grid must be its own slot's size    left: (40, 90)  right: (14, 30)
the agent was sized by something other than its own slot
                                                left: [(14, 30), (40, 90)]
                                                right: [(40, 90)]
a frame that changed nothing resized a pane     left: 24  right: 4
last_rect("<id>#shell")                         left: None
```

That is the flashing as a number — four resizes per frame, never settling — and
the agent being handed the shell's dimensions before its own.

Not verified: a real terminal or a remote host. The harness is offline by
construction, which is what makes the resize loop something CI can see instead
of something a human watches a terminal for.

## Attestation

```json
{
  "head_sha": "35b654c1e6caac95ce3e1b1833f831e1dd5a35ee",
  "base": "main",
  "base_sha": "53e4cc20841c39fc474931aed2bc39a2fe945821",
  "attested_at": "2026-09-21T10:59:51Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 3,
      "findings": 15,
      "fixed": 13
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
    },
    {
      "name": "documentation",
      "status": "passed"
    }
  ],
  "verdict": "passed"
}
```

— LeTuR's agent


LeTuR (commented):
Next: merge

The test I would have asked for is already here: "a frame that changed nothing resized a pane" pins the render-loop invariant directly — 24 resizes before, 4 after — and the inverted arrangement is the case that catches anything still assuming the agent owns `center`. Six of the twelve failing against the old kernel with the reported symptoms is the right kind of evidence.

The published side is in step: `thurbox.yml` and `thurbox.d.lua` both carry the new keying — `links` by surface, `content` by session with both panes — with no new field names, so selene and luals see the same shapes. Doubling `refresh_links` over surfaces stays behind the existing moved-and-`LINK_SCAN_INTERVAL` gate, so an idle shell costs the check and not a scan.

The bare-id-gets-the-agent change is named in the body and touches no bundled pane.

— LeTuR's agent

PR #1236, opened 2026-09-22
feat(ui): offer four layout presets, each validated by a demo video

LeTuR:
## Intent

Users choose an arrangement when they install: `classic` (unchanged, the default), `split-shell` (the shell in a pane below the agent), `focus` (the agent alone) and `ide` (sessions left, shell along the bottom, installed panes on the right). #1227 shipped the first two and was rolled back (#1233) because `split-shell` broke in real use while its tests passed. This PR brings the presets back after running them for real. Every defect that run found now has a test and a fix, and each preset has a demo video recorded against this branch.

## What Changed

- **The presets come back.** This branch reverts the rollback, then adds the stopped WIP's `focus` and `ide`. The install-time choice, `settings.toml`'s `layout`, `thurbox-cli layout list|set`, the settings row, the edited-layout backup, F-key toggles, `filled()` gating and the narrow fallback all behave as in #1227.
- **Defects found by running #1227's `split-shell` in a sandbox, each fixed with a test first:**
  1. `exit` in the shell pane left its last screen frozen. Keystrokes went nowhere and F8 couldn't bring it back. The classic Shell tab had the same bug. Now an exited shell is replaced the next time its surface paints.
  2. An agent pane edited before presets existed still offers its Shell tab, so the shell was drawn in two rects every frame and its pty flip-flopped between sizes. Now a surface paints into one rect per frame: the first.
  3. Narrowing (or shortening) the terminal while typing in the shell pane moved focus to the agent pane, on the agent's own view, so the next line went to the agent. The shell now keeps the keyboard both ways: it continues as the Shell tab, then hands back to the pane when there is room again.
  4. Switching to `split-shell` dropped hand-added plugin columns (a file tree, a queue). Now `split-shell` and `ide` stack installed panes in a right-hand column, which goes first below `three_panel_min_cols`.
  5. With the list hidden (F9), `thurbox-cli session focus` or a clicked notification was ignored until the list came back. The agent pane now takes the request.
  6. A preset named in `settings.toml` stayed out of force, with no message, while an edited `layout.lua` was in use. The first start now says so.
  7. Found while checking the recordings: output could go missing from a pane after a reflow until the next keystroke. The reader stamped "new output" before the parser had the bytes, and two chunks in the same millisecond produced the same stamp. Now it stamps only after feeding, and each chunk moves the stamp forward.
- **Found by review, fixed with tests.** An `exit`ed shell whose window still lives (a dropped ssh link) is reattached, not orphaned. It's replaced only when tmux says the pane is gone, which it checks with one bounded query. A failed open is retried with a backoff that starts when the open failed. `thurbox-cli layout set` and the settings row share one rule for which interface directory a switch may rewrite (`presets::may_switch`). The layout notice now leads the startup notices, because only the first one is shown.
- **Rebased onto #1235** (edited layouts keep working across upgrades). Its `panels.placed` shim is the real function again, `restore` keeps #1235's backup of an edited `layout.lua` and restores the preset that was last delivered, and the pinned `lib/` surface gains `panels.others` and `panels.starts`.
- **Demo videos.** `scripts/demo/layouts/record.sh` records the real TUI (asciinema + agg, driven through tmux, because VHS can't send F-keys). It uses a hermetic sandbox with a made-up agent and made-up sessions. The output is `media/layout-<preset>.{gif,mp4}` (about 200–250 KB each), embedded in `ui/README.md` and on the installation page.

```mermaid
flowchart LR
  pane["shell pane focused"] -- "narrow / short screen" --> tab["agent pane: Shell tab, focused"]
  tab -- "room again" --> pane
```

## Risk Assessment

- `classic` is unchanged byte for byte and stays the default.
- Anyone who moved to v2.32.1 had their `layout` key removed by that release's rollback, so they're on `classic`. This PR drops that one-time migration: `layout` is a valid key again.
- Two kernel changes reach every layout, classic included: an exited shell now reopens, and each output chunk's stamp moves only after its bytes are parsed.
- Under `focus`, installed panes stay closed until their own toggle opens them. So a pane with no toggle, like a file tree without one, isn't shown there. That's deliberate for a zen layout and documented.
- In the edited-agent-pane case (defect 2), the Shell tab wins and the shell pane stays blank rather than fighting it.
- The recorder and demo videos add about 1.7 MB to `media/`.
- A spawn from paint still runs on the loop thread, as it did in #1227 and as F8 does. A stalled remote link costs one blocking spawn per backoff step, capped at one a minute.
## Testing

- **Defects first.** Each defect above got a failing test before its fix: e2e tests on the real binary in `tests/tui_e2e.rs` (dead shell in the pane and in the tab, the keyboard across narrow and wide, ide F8, classic's Shell tab), plus `tests/layout_presets.rs` (third-party columns, focus requests with the list hidden, the handover both ways, the not-in-force notice). Kernel tests in `decoupling.rs` cover one rect per frame, reattach against replace, the retry and its backoff. `backend.rs` tests cover the output count and clock.
- **Each preset's placement, fallback and focus moves** are pinned in `tests/layout_presets.rs` and by one e2e test per preset.
- **Validated by running it.** Every preset was driven in a hermetic sandbox, with a made-up agent and made-up sessions, through the walk the videos show. I watched every recording frame by frame, at 120 columns and at the 100- and 70-column fallbacks. One more defect turned up only there (defect 7).
- **Review.** Seven rounds against `docs/REVIEW.md`: five, then two verification rounds on the last fixes, which is past the gate's five-round cap. It raised 13 findings, all fixed with tests, and the last round found nothing.
- **Gate.** `just lint` plus rustdoc `-D warnings`, and `cargo nextest run --all` (2968 passed), on the pushed head, after the rebase onto #1235. CI is green on that head, SonarCloud included. Its first run flagged a blocking sleep in an async test, fixed in the head.
- **Not verified here.** macOS, Windows/psmux (`install.Tests.ps1` is left to the Windows CI job), remote SSH hosts (the reattach path is covered by the kernel tests only), and the website linters (left to CI).

### Demo videos

| classic | split-shell |
|---|---|
| classic | split-shell |
| **focus** | **ide** |
| focus | ide |

Each clip chooses its preset in settings (F6 → `layout`), then walks through it: typing to the agent, F8 into the shell and back, switching session, F9, the installed `tasks` pane (split-shell and ide), and the terminal resized to 70 columns (and 100 for ide) and back.

### Try it yourself

1. Build the branch: `cargo build --bin thurbox --bin thurbox-cli`. To stay away from your live instance, use `scripts/dev/sandbox.sh --fresh` (it runs the build on PATH in an isolated profile).
2. Create two or three sessions, then press **F6**, move to the **layout** row, press **→** to pick a preset, **Ctrl+S**, **Esc**. The arrangement changes on the spot. `thurbox-cli layout set <classic|split-shell|focus|ide>` does the same from outside, and backs up an edited `layout.lua` to `layout.lua.bak` first.
3. **classic:** **F8** or **Ctrl+T** switches the agent pane between Agent and Shell. **F9** hides and shows the list.
4. **split-shell:** the shell is below the agent. **F8** moves focus into it and back. Select another session in the list and the shell follows. Type `exit` in the shell and a fresh one appears. Narrow the terminal below 80 columns while typing in the shell: it becomes the agent pane's Shell tab and keeps the keyboard. Widen again and focus goes back to the shell pane.
5. **focus:** only the agent shows. **F9** brings the list back. The shell is a tab (**F8**).
6. **ide:** list left, shell panel below the agent, and installed panes (for example `thurbox-cli plugin install examples/panes/tasks`) in a right-hand column. Narrow to 100 columns and the right column goes. Below 80 the agent is alone.
7. With your own edited `layout.lua` and `layout = "split-shell"` written by hand in `settings.toml`, the next start says in the message band that your edited layout is in use.

## Attestation

```json
{
  "head_sha": "34c509fe576dba51d15aed019d452146db108e87",
  "base": "main",
  "base_sha": "0b62deb9a044ae8dceff6fa2d0c5661301697b01",
  "attested_at": "2026-09-22T16:48:43Z",
  "steps": [
    {
      "name": "review",
      "status": "passed",
      "rounds": 7,
      "findings": 13,
      "fixed": 13
    },
    {
      "name": "lint",
      "status": "passed",
    },
    {
      "name": "test",
      "status": "passed",
    },
    {
      "name": "documentation",
      "status": "passed"
    },
    {
      "name": "ci",
      "status": "passed",
      "conclusion": "success",
      "run_url": "https://github.com/Thurbeen/thurbox/actions/runs/35753971103"
    }
  ],
  "verdict": "passed"
}
```

— LeTuR's agent


LeTuR:
The first run failed SonarCloud's reliability gate on one issue: a blocking `std::thread::sleep` in an async test (`decoupling.rs`). 34c509f replaces it with `tokio::time::sleep`. The gate now passes, with 0 open issues and 0 hotspots.

— LeTuR's agent

LeTuR:
IDE and split shelll are similar ?

LeTuR:
One layout could be enlarged session list with shell below

ui/layouts/ide.lua
```diff
@@ -0,0 +1,165 @@
...
+
+-- Below 20 rows the header goes first, the footer below 4.
+local HEADER_MIN_ROWS = 20
+local FOOTER_MIN_ROWS = 4
+
+local SEARCH_ROWS = 12
+
+-- The share of the centre column the shell takes, and the least it is worth
+-- showing: a shell a few rows tall shows a prompt and nothing it printed.
+local SHELL_PCT = 30
```
LeTuR: This file and `split-shell.lua` are the same code except for two numbers: sessions 20% vs 25%, shell 30% vs 35% (diffed with comments stripped). Since defect 4 gave `split-shell` the right-hand column, both presets arrange the same slots in the same places, which contradicts `presets.rs`'s "a preset earns its place". Suggest: drop one, or make `ide` different (for example the wider session list with the shell below it that the operator suggested).

— LeTuR's agent

ui/plugins/20_agent.lua
```diff
@@ -78,14 +93,54 @@ local function selected()
...
+---
+--- The list owns `store.selected` and writes it from its own render, which is
+--- also where it spends a `focus_session` request (a clicked notification,
+--- `thurbox-cli session focus`). A list that is not placed never renders — the
+--- `focus` preset starts it hidden, and F9 hides it anywhere — so this pane
+--- stands in for it: the request lands, and with nothing selected the list's
+--- first row is, the one its cursor would start on. Written only on a change,
+--- like the list's own writes, and the list adopts it through its `steer` the
+--- moment it is shown again.
+local function select_without_the_list()
```
LeTuR: The fixes for defects 3 and 5 live only in this pane, so any fork loses them. I installed thurbox-files' `20_agent.lua` (its README says to replace this one) under `focus`: the screen shows "No Session" with two sessions running, until F9 is pressed twice. Suggest a kernel or `lib/` fallback for `selected` when the list isn't placed, so a replaced agent pane keeps working.

— LeTuR's agent

src/coordinator/interface.rs
```diff
@@ -184,9 +191,48 @@ impl App {
...
+            self.toast(format!(
+                "layout {name} saved; layout.lua left as it is: {why}"
+            ));
+            return;
+        }
+        match thurbox::kernel::presets::apply(&self.ui_dir, name) {
+            Ok(applied) => {
+                self.refresh_sources();
+                self.reload_at = Some(Instant::now() + DEBOUNCE);
+                self.toast(match applied.backup {
```
LeTuR: This toast never reaches the screen. Right after it, the worker's "settings saved" and the reload's "settings reloaded" take the single status slot. I switched from the settings panel three times with an edited `layout.lua`, polling the screen every 100 ms: `.bak`, `.bak.2` and `.bak.3` were created, and not one was announced. Suggest reporting the backup after the reload, or refusing to switch with an edited file until the user confirms.

— LeTuR's agent

src/coordinator/publish.rs
```diff
@@ -23,6 +23,7 @@ impl App {
         // session is on screen — a stale set there would freeze the very
         // spinner it gates.
         self.terminals.sync_printing();
+        self.open_wanted_shells();
```
LeTuR: This runs `ensure_shell_pane` on the loop thread: one `pane_gone` round trip, then a spawn. It is triggered by a paint, not a key press. `docs/REVIEW.md` (`src/**`) says anything that touches the world runs on a worker. The backoff limits a dead remote host to one blocking spawn a minute. Is a worker handoff (like attach) out of scope for this PR? The operator should decide that explicitly.

— LeTuR's agent

docs/FEATURES.md
```diff
@@ -3121,6 +3121,14 @@ other panel toggles' F-keys.
   per session, so flipping to the shell on one does not flip it on the next.
   That is the pane's own state, not the kernel's — see the section below.
 - Input reaches the pane the surface names, `<id>#shell` or the bare id.
+- **`exit` is not the end of it**: a shell whose process ended is replaced by a
+  fresh one the next time its surface is painted, so the tab (or the shell pane
+  of the `split-shell` and `ide` layout presets) never sits on a frozen screen
+  that swallows keystrokes. `Session::ensure_shell_pane` treats an exited shell
```
LeTuR: Stale since review round 1: an exited shell whose pane tmux still has is reattached, not treated as none, and a failed ask is retried with a 2 s backoff doubling up to 60 s, not asked once. Suggest describing the reattach/replace rule and the backoff, or pointing to `Terminals::want_a_live_shell`.

— LeTuR's agent

ui/README.md
```diff
@@ -392,6 +393,78 @@ remove one.
...
+
+```bash
+thurbox-cli layout list              # which one is chosen, and whether layout.lua is edited
+thurbox-cli layout set split-shell   # switch; settings (F6) → layout does the same
+```
+
+An untouched `layout.lua` follows upgrades of the preset you chose. An edited one is
+yours: nothing overwrites it, and a switch you ask for moves it to `layout.lua.bak`
+first (`.bak.2` and on, never over an earlier backup) and tells you so. If
+`settings.toml` names a preset while an edited `layout.lua` is in use, the first
```
LeTuR: The notice appears on every start while the mismatch lasts: I restarted twice and got it both times, as `boot::layout_notice` calls it on each start. Also, nothing here says how to put a `layout.lua.bak` back. After a manual `mv`, the settings row still names the preset, and the only way to silence the notice is editing `settings.toml` to `classic`. Suggest a sentence on restoring a backup.

— LeTuR's agent

LeTuR (commented):
Not ready to merge yet: the four presets work as the PR describes when I run them, but two presets are one layout, forked agent panes lose the fixes, and a switch from settings moves your edited `layout.lua` aside without a word you can see. Findings, most severe first:

1. **`split-shell` and `ide` are the same arrangement.** Their code differs only in two percentages. Keep one, or make `ide` do something the others can't (see inline).
2. **Defects 3 and 5 are fixed in the bundled agent pane only.** With thurbox-files' fork of `20_agent.lua`, `focus` starts on "No Session" while two sessions run.
3. **Switching from the settings panel moves an edited `layout.lua` to `.bak` with no visible message.** The toast is overwritten at once by "settings saved" / "settings reloaded".
4. **A paint can open a shell on the render loop.** `docs/REVIEW.md` (`src/**`) forbids this. It is disclosed, and bounded by the backoff, but it is now automatic in every layout, classic's tab included. Rounds 3 to 7 of the review were all about this path (reattach, backoff, failure timing). The operator decides.
5. **The `settings.toml` key and `layout.lua` can disagree** (operator's question, below).
6. **Docs.** `FEATURES.md` describes the pre-review shell logic. `ui/README.md` says the notice appears on "the first start", but it appears on every start. Nothing says how to put a `.bak` back. The body says v2.32.1 rolled back; that was v2.32.2. "classic byte for byte" holds for `layout.lua` only: classic users also get `25_shell.lua`, the first-run notice and the new agent-pane behaviour.
7. **Scope.** The output count (defect 7) and the dead-shell reattach/backoff are `fix(core)` changes that reach every layout. Landing them first, as their own PR, would leave this one about presets. Separately, the tests-first commit 9cb483209 no longer compiles after the rebase (its `decoupling.rs` harness needs a type from a later commit).
8. nit: `record.sh` needs agg and asciinema, which `flake.nix`'s `demoTools` doesn't provide, and `WORK` defaults to `mktemp` under `/tmp`. I didn't re-run it (no agg here), so its reproducibility is unverified.

**What I ran.** I built the head (34c509f) and drove each preset in a hermetic sandbox: its own HOME, XDG dirs and tmux sockets, a made-up agent, two sessions. Switching works from settings, from `thurbox-cli layout set` and by editing `settings.toml`. In the shell pane, F8 moves focus in and out, the shell follows the selected session, and `exit` gives a fresh shell. The keyboard stays on the shell at 70 columns, at 18 and 12 rows, and back again. `ide` drops the right column at 100 columns and leaves the agent alone at 70. F9 and `session focus` work with the list hidden. Hand-added columns appear in the right column. I watched all four videos: they show their preset, and nothing private is in any frame.
**Defect claims.** 1–4: 8 tests fail at 9cb483209 for the right reason (after a one-line compile fix) and pass at head. 5 and 7: the tests fail when I switch the fix off at head. 6: its test cannot compile before the fix, so I took it on a live run. Live, I reproduced 1, 3, 4, 5 and 6. For 2 and 7 I rely on the tests, plus the fork run for 2 (the Shell tab wins; the shell pane shows a "terminal surface" placeholder). I didn't run the full suite; CI is green. The media adds 1.7 MB, in line with `media/` (19 MB, excluded from review).

**Operator question: does switching break a custom layout? Settings, or the Interface tab?** Nothing was lost or silently overwritten. A column I added to classic and a fully custom arrangement were each kept:
- Hand-editing `settings.toml` leaves the edited file alone, and every start says so. With a custom layout that has no `status` slot, nobody sees that line.
- The settings panel and the CLI each move the edit to `layout.lua.bak[.N]`. Only the CLI tells you (finding 3).
- Across an upgrade, an untouched old preset is refreshed to the new copy and an edited one is kept.
- Putting a backup back is a manual `mv`, and it is documented nowhere. Afterwards, the settings row and `layout list` still name a preset that isn't on screen, and the notice repeats every start. The only verb that would silence it (`layout set`) moves your file aside again.

That mismatch is the real cost of the settings key: it holds a second copy of what the file already says. I'd go with the Interface tab.
- `layout.lua` is already a row there, with bundled/edited state and restore-with-backup.
- A preset picker on that row would show "edited" at the moment you switch, and could ask before moving your file.
- Delivery can tell which preset to refresh from the manifest: `bundled::delivered_preset` already does exactly that.
- Cost: delete `Settings.layout`, its core row, `not_in_force` and the start notice (mostly deletions); add one action and a preset label to the Interface modal; keep `thurbox-cli layout set` as the scripted verb, minus its `settings.toml` write; rewrite the settings-row tests and the CONFIG/PLUGINS/README sections.

— LeTuR's agent

